PR #17, opened 2019-07-26
 Modify the README content and resolve the title display exception

rabbityo:
Modify the README content and resolve the title display exception


yangzhiyue:
thanks for contributing

PR #18, opened 2019-07-26
Modify hyperlink

rabbityo:
Modify hyperlink


yangzhiyue:
thanks for contributing

PR #30, opened 2019-07-30
Fix user manual 404 error

leeebai:
Found that the path to the user manual in the read.me file is incorrect.


wushengyeyouya:
thanks!

PR #32, opened 2019-08-01
remove useless infomation

patinousward:
remove useless infomation in class


wushengyeyouya:
Thanks for contributing!

PR #33, opened 2019-08-02
correct stop_all.sh to stop-all.sh

leisore:
stop_all.sh should be stop-all.sh in script install.sh


wushengyeyouya:
Thanks for contributing!

PR #34, opened 2019-08-02
modify linkis_ddl.sql, fix 'linkis_udf_tree' sql bug

liangqilang:
`update_time` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00'


wushengyeyouya:
Thanks for contributing!

Issue #67, opened 2019-10-08
job configuration Graininess smaller


patinousward:
**Is your feature request related to a problem? Please describe.**
when i submit spark job ,I found that i can't setting configuration for diffrent job(in a short time),console is good,but i am tied of  change the setting frequently，and i hope the configuration Graininess smaller

**Describe the solution you'd like**
i hope i can setting some start up  configuration  when i run a job(or a history job) just like variable ,i hope the configuration can overwirte the console 

**Describe alternatives you've considered**
i think the Efficiency of persist in the script file may be higher than database

**Additional context**
nothing

PR #81, opened 2019-10-14
Modify script resolution

patinousward:
#67 
Modify script resolution
1.Does not affect the original function
2.provide interface for #67


patinousward:
Modify script resolution

PR #88, opened 2019-10-21
if hadoop version is 2.8.+ it is necessary to aviod the hdfs schema c…

xinda77:
…ann't support  error when linkis initialize the workspace

### DIFF
--- a/core/cloudModule/pom.xml
+++ b/core/cloudModule/pom.xml
@@ -201,6 +201,10 @@
                     <groupId>org.springframework</groupId>
                     <artifactId>spring-context</artifactId>
                 </exclusion>
+                <exclusion>
+                    <artifactId>jersey-server</artifactId>
+                    <groupId>org.glassfish.jersey.core</groupId>
+                </exclusion>
             </exclusions>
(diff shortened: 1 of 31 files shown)


chaogefeng:
@xinda77 Hello, can you describe your problem in detail?

xinda77:
> @xinda77 Hello, can you describe your problem in detail?

i build the project with CDH6.1.1 and i cann't get the workspace ,so i check the publicservice's log got the error  "org.apache.hadoop.fs.UnsupportedFileSystemException: No FileSystem for scheme "hdfs""
the error is caused by the lack of a scheme for hdfs, which means that FileSystem lacks the implementation of the corresponding hdfs.
this is because the org.apache.hadoop.hdfs.DistributedFileSystem class
In hadoop2 in hadoop-hdfs-*.jar
But it was migrated to  hadoop-hdfs-client-*.jar in hadoop3
so we should set the class and the dependency when we used in hadoop3

peacewong:
Hello, can this pr be submitted separately, I see a lot of changes

PR #91, opened 2019-10-22
Compile Linkis based on hadoop 3.1.1

Just-do-it-Fan:


sargentti:
We are discussing a complete solution about muti-version support for the back-end engines.

wushengyeyouya:
Excuse me, Can you help split this PR into 2? One for the Yarn capacity scheduler, another for the compatibility of hadoop3.
And it is better if you can submit the new 2 PRs to dev-0.11.0.

PR #99, opened 2019-11-01
impala引擎

liangqilang:
扩展impala引擎，采用Thrift协议与coordinator通讯，实时获取日志以及进度条，支持kill操作。


peacewong:
Thank you very much. 
There are several amendments
1.impala should be placed in the definedEngines directory
2. Is the code related to the beginning of the com.cloudera and org.appache packages in impala included in the package? Don't use the code
3.The thrift file license in thrift needs to be modified to the same

Issue #66, opened 2019-09-30
Scala script 1 created method, scala script 2 can be called directly, sql call can not recognize the udf function registered by script 1


zhanghaicheng1:
1, sql can not recognize the udf created by scala;
2. The script udf (method) created by scala is not registered and can be called directly.

Issue #109, opened 2019-11-18
Linkis-ujes-client module feature optimization


peacewong:
**Is your feature request related to a problem? Please describe.**
1. The return of jobLogResult should be an array
2. JobExecution should support the use of strings to determine engineType and RunType
3. Modify the request publicservice to the corresponding service name

PR #110, opened 2019-11-18
Fix this issue that the way ip:port exists when Engine gets the default path of the result set

### LINKED ISSUE #108: The way ip:port exists when Engine gets the default path of the result set
**Describe the bug**
The ip:port method exists when the Engine obtains the default path of the result set. Because the path has the symbol ':', the hdfs creation path fails.

**Additional context**
Need to modify the way the default path is obtained

peacewong:
Modify how the engine gets the default result set path
Closes #108,#66


peacewong:
Closes #108,#66,#109

Issue #112, opened 2019-11-19
metadata module enhancement


patinousward:
**Is your feature request related to a problem? Please describe.**

I want to know more about the table metadata in mysql，hbase...  not just hive

**Describe the solution you'd like**
Expand the metadata module ，It has several tables to record more metadata message

PR #113, opened 2019-11-19
metadata module enhancement

patinousward:
#112  
1.It provides more information about hive ，such as table size，number of files，lifecycle and so on
2.does not affect the original foreground interface
3.Later, it can be integrated with msyql，hbase...


patinousward:
mix some other commit message, close it

PR #125, opened 2019-11-21
metadata module enhancement

patinousward:
#112 
1.It provides more information about hive ，such as table size，number of files，lifecycle and so on
2.does not affect the original foreground interface
3.Later, it can be integrated with msyql，hbase...


chaogefeng:
@peacewong @wushengyeyouya  I have already communicated to the author,it's ok to merge into the master, thank you very much for your contribution. @patinousward

Issue #120, opened 2019-11-19
workspace module enhancement


patinousward:
**Is your feature request related to a problem? Please describe.**

Workspace lacks file interaction with BML

**Describe the solution you'd like**
1.provice new  http interface to save or parse script file in BML
2.provice workspace-http module ,other modules can use it to access BML ，and get script

peacewong:
This issue has been fixed

PR #130, opened 2019-11-21
workspace module enhancement

patinousward:
#120 
1.provice new http interface to save or parse script file in BML
2.provice workspace-http module ,other modules can use it to access the HTTP interface of workspace more easily


chaogefeng:
@peacewong @wushengyeyouya I have already communicated to the author,it's ok to merge into the master, thank  for your contribution. @patinousward

PR #144, opened 2019-11-24
Linkis-gateway: fix ldap dn chinese garbled code

nimuyuhan:
issues: https://github.com/WeBankFinTech/Linkis/issues/143
Fix the bug  ldap dn chinese garbled code.


peacewong:
Thanks, what do you want to fix is the problem that the configuration file Chinese will be garbled?

PR #349, opened 2020-04-16
fix some bugs

wForget:
modify the implementation of Base64 #320
add ldap userNameFormat
modify HiveDriverProxy log
fix a VariableParser bug
add yarn application id log in spark engine #323
fix jdbc engine executor write result set bug. #361


Alexkun:
Thanks.

Alexkun:
Thanks.

Issue #403, opened 2020-05-29
some code quality alerts in bml and publicservice module


patinousward:
1.The number of log print information placeholders and variables is inconsistent
2.Some streams are not closed
3.Singleton classes that may cause problems
4.String should not use in synchronized（）

yangzhiyue:
thanks it is fixed in #404

PR #404, opened 2020-05-29
fix some alerts

patinousward:
#403 

Solve this problem


storage/storage/src/main/java/com/webank/wedatasphere/linkis/storage/excel/ExcelStorageReader.java
```diff
@@ -39,7 +39,7 @@
         } else {
             res = XlsxUtils.getBasicInfo(in, file);
         }
-        if(res == null && res.size() < 2){
+        if(res.size() < 2){
```
Alexkun, line 42: Should check if null in case of nullpoint exception

patinousward: In fact, it would't be null

wushengyeyouya:
Please resubmit this PR to dev-0.11.0, thanks very much!

PR #457, opened 2020-06-19
Python Engine BML zip file support

houjunxiong:
**Is your feature request related to a problem? Please describe.**
When we use python engine, we want to add a series of user own defined classes and functions in our own python *.py files, but there is no way to add it to the python engine now.

我们在使用Pyhton引擎时，想要使用自定义的一系列自定义的类和函数（定义在一系列*.py文件中），但是现在的Python引擎无法满足该需求。

**Describe the solution you'd like**
We have solved the problem. Python engine can import bml resources like zip files which contain *.py files， and the bml resources are defined in "entrance/execute" interface params.configuration.runtime.resources, the struct is "[{'resourceId': ***, 'version': ***, 'filename': ***}]", when the python engine starts to execute some code, it will download the zip resources from bml service and import them into the python environment.

当前，我们已经解决了该问题。我们在"entrance/execute"接口中定义了一个参数"params.configuration.runtime.resources"，参数值为一个bml资源数组，结构为"[{'resourceId': ***, 'version': ***, 'filename': ***}]"，当Python引擎开始执行一段代码之前，将从BML服务下载对应的资源文件，并将它们导入Python环境中，从而实现使用自定义类和函数的功能。

**Describe alternatives you've considered**
None

**Additional context**
None

### DIFF
--- a/bml/bml-engine-hook/src/main/scala/com/webank/wedatasphere/linkis/bml/conf/BmlHookConf.scala
+++ b/bml/bml-engine-hook/src/main/scala/com/webank/wedatasphere/linkis/bml/conf/BmlHookConf.scala
@@ -1,5 +1,6 @@
 package com.webank.wedatasphere.linkis.bml.conf
 
+import com.webank.wedatasphere.linkis.bml.utils.BmlHookUtils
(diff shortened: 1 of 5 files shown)


peacewong (approved):
Thanks!

Issue #469, opened 2020-07-04
ldap tool class add method "pass in useName and key to get the value"


waterWang:
The public method getValueByName is added to the LDAPUtils class, which is used to obtain the corresponding attribute value in ldap according to the transmitted uid (useName). For example, if key=mobile is passed in, the mobile information in ldap is obtained

LDAPUtils类中添加公共方法getValueByName，作用：根据传输的uid（useName）获取ldap中对应的属性值，例如传入key=mobile，则获取ldap中mobile信息

PR #470, opened 2020-07-04
#469

waterWang:


wushengyeyouya:
Excuse me, it is more wonderful if you can give a detail usage example of this enhancement.

PR #473, opened 2020-07-06
Feature#472

waterWang:


wushengyeyouya:
This PR brings lots of useful improvements, it seems like to be better if log content optimization, log interface optimization, and code indentation optimization are divided into several PRs.

PR #483, opened 2020-07-29
Dev 0.11.0 add some commits

wForget:
fix yarn own queues bug
add auto get acl queues config
remove org.scalactic.Pass dependents
fix regexp StackOverflowError


peacewong (approved):
Thanks!

PR #519, opened 2020-12-01
linkis on k8s and api service

brianzhangrong:


yangzhiyue (approved):
thanks for this pr

Alexkun (approved):
Thanks.

peacewong (approved):
Thanks, linkis on k8s is an excellent feature

Issue #489, opened 2020-08-19
[Bug][SSO] The problem of user cookie loss after single sign-on access


linweijiang:
After SSO, web returns to the homepage, if you continue to click on other pages, the authentication failure will occur. After checking the page request details, it is found that there is no bdp-user-ticket-id cookie

单点登录成功返回首页后，继续点其他页面会出现认证失败问题，查看页面请求详情后发现没有 bdp-user-ticket-id cookie

wushengyeyouya:
Thanks very much!

PR #532, opened 2020-12-16
#489  fix loss of user cookie when login by sso

linweijiang:
#489 fix loss of user cookie when login by sso


peacewong (approved):
Thanks

PR #534, opened 2020-12-17
spark引擎中加入streaming依赖

77954309:
spark引擎中加入streaming依赖
image


wushengyeyouya:
Thanks for your contribution，but I'm sorry to advice that, this PR will introduce lots of additional dependencies, so we have to refuse it.

PR #537, opened 2020-12-17
modify name of linkis_udf_tree to varchar(200)

hcl3039359:
modify name of linkis_udf_tree to varchar(200)


peacewong (approved):
Thanks for your fix

PR #545, opened 2020-12-29
change version 0.10.0 to 0.11.0 and fix es entrance

wForget:


peacewong (approved):
Thank you, it is necessary to modify the version

PR #553, opened 2021-01-14
remove getValueAndRemove method from linkis-common

wForget:
remove getValueAndRemove method from linkis-common


peacewong (approved):
Thanks!

PR #556, opened 2021-01-26
[Fix Bug]Don't filter out Visualis-share URLs

silly-carbon:

This BUG is fixed in this PR.


wushengyeyouya:
I think this optimization is great. It would be better if you can help to pass it according to regex matching and make regexes read from configuration file.

silly-carbon:
@wushengyeyouya

silly-carbon:
@wushengyeyouya Sorry for the late update.
I have made these changes for what you mentioned earlier, which was great.

Mgthanlwin (approved):

peacewong:
Hello, there is a conflict between the 0.X and 1.0 codes, it is recommended to reopen this PR from the dev-1.1.0 branch

PR #563, opened 2021-02-04
Standardize the directory structure of linkis

peacewong:
Standardize the directory structure of linkiss, which is divided into three major modules: public enhancement, computing governance, and microservices。


wushengyeyouya (approved):
Thanks for optimizing the structure of linkis modules.

Issue #576, opened 2021-03-09
Add a new module EngineConnPlugin for the Linkis1.0 architecture to simplify the user's implementation of a new Linkis underlying computation storage engine.


wushengyeyouya:
Add a new module EngineConnPlugin for the Linkis1.0 architecture. This module is defined to simplify the user's implementation of a new Linkis underlying computation storage engine.  It provides the following capabilities:

1. A separate service that provides interfaces for LinkisManager to  obtain the resources when creating an EngineConn.

2. Provides interfaces for EngineConnManager to prepare the start-up argument information of EngineConn after ECM receives the EngineConnBuilder's requests from LinkisManager.

3. Be able to load the dependency jars required by the computation storage engine in real time, so those engines, which have been integreted to Linkis don't need to be pre-installed in EngineConnManager machines.

4. Be able to load the corresponding EngineConnPlugin on demand.

PR #595, opened 2021-03-10
Add a new module EngineConnPlugin for the Linkis1.0 architecture to simplify the user's implementation of a new Linkis underlying computation storage engine.

wushengyeyouya:
Related issues: #576
Add a new module EngineConnPlugin for the Linkis1.0 architecture. This module is defined to simplify the user's implementation of a new Linkis underlying computation storage engine.

### Brief change log

1. A separate service that provides interfaces for LinkisManager to  obtain the resources when creating an EngineConn.

2. Provides interfaces for EngineConnManager to prepare the start-up argument information of EngineConn after ECM receives the EngineConnBuilder's requests from LinkisManager.

3. Be able to load the dependency jars required by the computation storage engine in real time, so those engines, which have been integreted to Linkis don't need to be pre-installed in EngineConnManager machines.

4. Be able to load the corresponding EngineConnPlugin on demand.


yangzhiyue (approved):
thanks for this pr

yangzhiyue (approved):
thanks for this pr

Issue #577, opened 2021-03-09
Split EngineReceiver of Linkis0.x into service and encapsulate executor


JinyangRao:
This module splits the EngineReceiver of Linkis0.x into multiple services, and decouples the service and listener. These services are scattered into multiple exeuctor modules, which are freely selected by specific types of engines according to needs, and the details of each are as follows:
1. The core module of executor, including the entity class of excutor and the top-level abstract interface, and the basic service of LabelService.

2. Accessable-executor module, including lock and heartbeat services.

3. The resource-executor module, including resource management services

4. Callback-service module, including callback AM and EM services.

PR #606, opened 2021-03-11
Split EngineReceiver of Linkis0.x into service and encapsulate executor

JinyangRao:
Related issues: #577  
This module splits the EngineReceiver of Linkis0.x into multiple services, and decouples the service and listener. These services are scattered into multiple exeuctor modules, which are freely selected by specific types of engines according to needs, and the details of each are as follows:

### Brief change log

1. The core module of executor, including the entity class of excutor and the top-level abstract interface, and the basic service of LabelService.
2. Accessable-executor module, including lock and heartbeat services.
3. The resource-executor module, including resource management services.
4. Callback-service module, including callback AM and EM services.


leeebai (approved):
LGTM.

Issue #604, opened 2021-03-11
Add the ability of microservice tagging, and support the tagging of each microservice


liuyou2:
To provide the ability of tagging for the micro services of linkis, the specific function points are as follows:

1. Support to increase or decrease tags for related micro services of linkis

2. A unified storage layer is abstracted to allow the tag information of microservice to be stored in other storage media such as database / service registration discovery service

3. Provide a unified client and support crud for the tag information of microservice through RPC

4. The metadata information of Eureka can be registered as a label.

PR #607, opened 2021-03-11
Add the ability of microservice tagging, and support the tagging of each microservice

liuyou2:
Related issues: #604
To provide the ability of tagging for the micro services of linkis, the specific function points are as follows:

### Brief change log
1. Support to increase or decrease tags for related micro services of linkis

2. A unified storage layer is abstracted to allow the tag information of microservice to be stored in other storage media such as database / service registration discovery service

3. Provide a unified client and support crud for the tag information of microservice through RPC

4. The metadata information of Eureka can be registered as a label.


wushengyeyouya (approved):
LGTM.

Issue #608, opened 2021-03-11
Enhance scritps for starting and shutting down services, optimize steps of construction


Alexkun:
These modules contain scripts of management, configuration and construction module, as follows:
1. bin - The install scripts mainly implement variable substitution
2. sbin - Provide linkis-daemon script to start and shutting down  services
3. conf - Abstract the common configuration, and private configuration is placed in specific structure. 
4. assembly - Optimized the steps of building package in assembly

PR #627, opened 2021-03-11
Enhance scritps for starting and shutting down services, optimize steps of construction.

Alexkun:
Related issues: #608
Enhance scritps for starting and shutting down services, optimize steps of construction.

### Brief change log
These modules contain scripts of management, configuration, db and construction module, as follows:
1. bin - The install scripts mainly implement variable substitution
2. sbin - Provide linkis-daemon script to start and shutting down  services
3. conf - Abstract the common configuration, and private configuration is placed in specific structure
4. db - Scripts on building databases during installation
5. assembly - Optimized the steps of building package in assembly


peacewong (approved):
LGTM.

PR #682, opened 2021-03-14
Modify the return of the job submit method

peacewong:

Modify the return of the job submit method

### Brief change log

1. Modify the return response submitted by the job


wushengyeyouya (approved):
LGTM.

PR #687, opened 2021-03-18
Fixed the problem that the jdbc component in the workflow failed to execute because the jdbc attribute information could not be read.

### LINKED ISSUE #686: The jdbc component of Linkis0.11.0 reports an error <ERROR DWCException{errCode=70012, desc='jdbc url or username or password may be null at least'>  when executed in the workflow
I used the jdbc component in the workflow，and The jdbc parameters have been configured in the console

I have configured the parameters of jdbc in the console, and the test sql can be executed successfully in the jdbc component, as shown in the figure below.
image

However, the following error will be reported when the jdbc component is executed in the workflow.
image

The error message is <ERROR DWCException{errCode=70012, desc='jdbc url or username or password may be null at least'>
In fact, the jdbc parameters have been configured.
image

emerkfu:
The error message is <ERROR DWCException{errCode=70012, desc='jdbc url or username or password may be null at least'>.

Fixes #686


peacewong (approved):
LGTM.

JohnTao:
nice

PR #706, opened 2021-04-08
fix bmlclient bug

### LINKED ISSUE #705: Bml client filesystem memory leak
bml client causes filesystem memory leak

the filesystems should be closed at once when after using

yangzhiyue:
close #705 

I fixed this bmlclient bug, and make the bmlclient one for everyone


peacewong (approved):
LGTM.

PR #729, opened 2021-05-14
Solve the problem that cookies does not take effect after login, and cookies added to Action is not delivered to gateway by the HttpClient.

wushengyeyouya:
Solve the problem that cookies does not take effect after login, and cookies added to Action is not delivered to gateway by the HttpClient.

## Brief change log
Optimize AbstractHttpClient.


yangzhiyue (approved):
thanks for this pr

PR #749, opened 2021-05-28
Add a new linkis-io-file-client module, modify some comments and add some license

JinyangRao:
This PR adds the following content:
1. Add a new linkis IO file client module.
2. Modify some comments and add some license.


owenlovemika (approved):
LGTM.

PR #780, opened 2021-06-05
Fix the value conversion problem of RM module in capacity scheduler

### LINKED ISSUE #778: The resourceManager module has a compatibility bug when Yarn is in capacity scheduler mode
**Describe the bug**
The resourceManager module has a wrong conversion when Yarn is in the capacity scheduler mode to obtain the maximum available resources of the Yarn queue.

**Expected behavior**
There is an incorrect conversion from JDouble to JDecimal, and the error log is as follows
```
Caused by: java.lang.ClassCastException: org.json4s.JsonAST$JDouble cannot be cast to org.json4s,JsonAST$JDecimal
```

peacewong:

Compatible with value conversion issues when Yarn is set to capacity scheduler.

### Brief change log

Modify the direct value conversion of the maxEffectiveHandle method

Closes #778


wushengyeyouya (approved):
LGTM.

PR #785, opened 2021-06-06
Issue7.Issue8 modify trycatch/submit sql

77954309:
Issue7. Linkis代码中，直接调用try catch的，要改为调用Utils.tryCatch
Issue8. 用户提交的sql如果select与*没有空格，就会导致不会自动加上limit 5000，可能会造成全表搜索，导致结果集很大，占用很大的hdfs空间

Issue7. In the Linkis code, if you call try catch directly, you need to call Utils.tryCatch instead
Issue8.If the sql submitted by the user does not have a space between select and *, limit 5000 will not be automatically added, which may cause a full table search, resulting in a large result set and taking up a lot of HDFS space

### DIFF
--- a/linkis-commons/linkis-message-scheduler/src/main/java/com/webank/wedatasphere/linkis/message/context/MSContextInstance.java
+++ b/linkis-commons/linkis-message-scheduler/src/main/java/com/webank/wedatasphere/linkis/message/context/MSContextInstance.java
@@ -16,9 +16,12 @@
 
 package com.webank.wedatasphere.linkis.message.context;
 
+import com.webank.wedatasphere.linkis.common.utils.Utils;
 import com.webank.wedatasphere.linkis.message.utils.MessageUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import scala.runtime.AbstractFunction0;
+import scala.runtime.AbstractFunction1;
 
 /**
  * @date 2020/9/17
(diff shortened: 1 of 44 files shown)


linkis-commons/linkis-message-scheduler/src/main/java/com/webank/wedatasphere/linkis/message/context/MSContextInstance.java
```diff
@@ -33,16 +36,25 @@ public static MessageSchedulerContext get() {
         if (context == null) {
             synchronized (MSContextInstance.class) {
                 if (context != null) {
-                    try {
-                        MessageSchedulerContext bean = MessageUtils.getBean(MessageSchedulerContext.class);
-                        if (bean != null)
+
+                    context = Utils.tryCatch(new AbstractFunction0<MessageSchedulerContext>() {
```
peacewong, line 40: Java code recommends using Try Catch directly

linkis-computation-governance/linkis-entrance/src/main/scala/com/webank/wedatasphere/linkis/entrance/interceptor/impl/Explain.scala
```diff
@@ -145,7 +145,8 @@ object SQLExplain extends Explain {
     }
     //如果一段sql是 --xxx回车select * from default.users，那么他也是select语句
     val realCode = cleanComment(code)
-    realCode.trim.split("\\s+")(0).equalsIgnoreCase("select")
```
peacewong, line 148: This bug fix suggests a separate PR

PR #787, opened 2021-06-06
[Fix-786][Bug] Fix the package conflict problem caused by hive version upgrade.

### LINKED ISSUE #786: [Bug] The hive version upgrade has a package conflict problem
**Describe the bug**
There are multiple package conflicts after the hive version is upgraded to 2.3.

**Expected behavior**
Because hive EC will add public-module to the classPath at startup, it is recommended to remove this conflicting package.

peacewong:

In order to solve the problem of multiple package conflicts caused by hive upgrade.

### Brief change log

Modify the assembly file to exclude publicly dependent packages.

Closes #786


yangzhiyue (approved):
thanks for this pr

PR #789, opened 2021-06-06
'utf-8'、'iso*' constant added

lemon-simple:
Fixes #issue #782 
Change to unified “utf8、ISO8859-1” and some other words


wushengyeyouya (approved):
LGTM.

PR #795, opened 2021-06-08
Issue8 submit sql

77954309:
If the sql submitted by the user has no space between select and *, limit 5000 will not be automatically added, which may cause a full table search, resulting in a large result set and taking up a lot of HDFS space


yangzhiyue (approved):
thanks for this pr

yangzhiyue (approved):
thanks for this pr

PR #805, opened 2021-06-10
issue7 Utils.tryCatch

77954309:
In the Linkis scala code, if you call try catch directly, you need to call Utils.tryCatch instead


yangzhiyue (approved):
thanks for this pr

PR #806, opened 2021-06-10
LinkisManager Client

77954309:
askEngine: engine access service, if there is reuse, there is no creation
createEngine：engine creation
stopEngine：release engine
reuseEngine：reuse engine, only existing engine can be reused
listUserEngines: get basic information of EngineNode through user, including metric
listEMEngines : get basic engine information through em (mainly instance information), including metric


yangzhiyue (approved):
thanks for this pr

PR #809, opened 2021-06-17
Update engine conn name

peacewong:

Because engineConn applicationName is changed to linkis-cg-engineconn, modify Alisa's default engineConn name.

### Brief change log

1. Adjust the alias of the default engineConnAliasLabel as a configuration parameter


wushengyeyouya (commented):
LGTM.

wushengyeyouya (approved):

PR #822, opened 2021-06-21
Shutdown all executors instead of only close the default one For multi-executor EngineConns

### LINKED ISSUE #816: When EngineConn process exits, the shutdown hook only closes the default Executor, not all Executors.
When EngineConn process exits, the shutdown hook only closes the default Executor. It should close all executors in the senario with multiple Executors (for example Spark).
For multi-executor EngineConns (such as SparkEngineConn), some executors are not shutdown properly so Yarn application cannot be closed normally.

sargentti:
For multi-executor EngineConns (such as SparkEngineConn), shutdown all executors instead of only close the default one.
Close #816 

### Brief change log
. Shutdown all executors in executorShutDownHook for multi-executor EngineConns.


peacewong (approved):
LGTM.

Issue #851, opened 2021-07-08
The version of the jackson-core-asl package is incorrect after linkis1.0 is compiled with the CDH version


labulabuduo:
hadoop.version was modified to 2.6.0-cdh5.16.1
After compiling and installing, The version of the jackson-core-asl package in the lib/linkis-commons/public-module directory is 1.8.10，It is different from the 1.9.13 version after the community-provided package is installed, which will cause the problem of failure to save any parameter configuration in the computing management console。

Exclusion of org.codehaus.jackson in the linkis-hadoop-common/pom file does not take effect when using wildcard *, so the default dependency jackson-core-asl-1.8.10 of the hadoop-common-2.6.0-cdh5.16.1 version is obtained when compiling。

PR #856, opened 2021-07-09
Fix "The version of the jackson-core-asl package is incorrect after linkis1.0 is compiled with the CDH version"

labulabuduo:
#851


leeebai (approved):
LGTM.

leeebai:
Thank you for your contribution to Linkis~

PR #857, opened 2021-07-12
new function, tenant label will not be delete when EM node restarted

### LINKED ISSUE #848: The tenant label of ECM should be retained to faciliate restart.
It is recommended that the tenant label of ECM should be retained. Therefore, considering adding the function of not cleaning up specific tags when the instance exits.

leeebai:
fixed #848
1. Tenant label will not be delete when node restarted
在节点重启时，租户标签将不被清理
2. You can determine which labels can be retained when the node restarts by setting LabelManagerConf.LONG_LIVED_LABEL
可以通过设置LabelManagerConf.LONG_LIVED_LABEL来决定哪些label可以在节点重启时被保留


owenlovemika (approved):
LGTM.

Issue #864, opened 2021-07-14
Optimize the default field length of some tables, and the compatible content is too long


misaki73:
Some fields of some tables are too long, causing compatibility problems.

PR #865, opened 2021-07-14
Optimize the compatibility problem caused by the default field is too long. fiexed #864 

misaki73:
fiexed #864 .
Optimized the problem that the fields in the following table are too long to solve the compatibility problem：
linkis_ps_configuration_config_value
linkis_ps_cs_context_id
linkis_ps_instance_label_relation
linkis_cg_manager_engine_em
linkis_cg_manager_label_service_instance


peacewong (approved):
LGTM

Issue #863, opened 2021-07-14
Spark native parameters need to be optimized for support, as well as support for setting the Python version


HmhWz:
Spark native parameters need to be optimized for support, as well as support for setting the Python version

PR #866, opened 2021-07-14
fix Spark native parameters need to be optimized for support, as well as support for setting the Python version #863

HmhWz:
fix Spark native parameters need to be optimized for support, as well as support for setting the Python version #863


peacewong (approved):
LGTM.

Issue #872, opened 2021-07-14
The cache cannot be refreshed in real time


FinalTarget:
The data in the cache cannot be changed in real time during the access process, and the data is still the previous version

PR #876, opened 2021-07-14
[#872]Optimize RPC caching function and HDFS FileSystem to support caching

FinalTarget:
The data in the cache cannot be changed in real time during the access process, and the data is still the previous version


peacewong (approved):
LGTM.

PR #878, opened 2021-07-14
Optimize the sequence of engine bml execution hook

workplease:
## What is the purpose of this request?
- Optimize the sequence of engine bml execution hook and fix the problem of not being able to read resource files  
- Support for real-time configuration updates


peacewong (approved):
LGTM

PR #895, opened 2021-07-15
Some bugs in linkis-entrance module need to be fixed.

### LINKED ISSUE #884: Some bugs in linkis-entrance module
**Describe the bug**
There are some bugs in linkis_entrance module:
① Job final status may be written by status from engine
② Job kill will not work with big job that has many sub jobs
③  Some error messages are not printed correctly
④  Method of Killjob doesn't work for more than one entrance server
⑤  Hard to get request path for job in case of multi entrance server

Alexkun:
Close #884
There are some bugs in linkis_entrance module:
    ① Job final status may be written by status from engine
    ② Job kill will not work with big job that has many sub jobs
    ③ Some error messages are not printed correctly
    ④ Method of Killjob doesn't work for more than one entrance server
    ⑤ Hard to get request path for job in case of multi entrance server

### Brief change log

    ① Changed version to 1.0.1
    ② Refactored linkis-entrance module to fix bugs above


peacewong (approved):
LGTM.

PR #899, opened 2021-07-15
Fix the bug of null pointer in log printing

wuliyong225:
Fix the bug of null pointer in log printing, and change 'message. ToString' to 'message'


peacewong (approved):
LGTM.

Issue #917, opened 2021-07-26
JDBC engine connection url is not configured


Adamyuanyuan:
JDBC engine connection url is not configured

PR #923, opened 2021-07-27
fix jdbc url is not configured

FireFoxAhri:

Fix a bug that JDBC engine connection url is not configured.

Related issues: #917.

### Brief change log
- Add jdbc url when it is null.

### Verifying this change
This change is a trivial rework / code cleanup without any test coverage.  

### Documentation
- If yes, how is the feature documented? (not applicable)

### DIFF
--- a/linkis-engineconn-plugins/engineconn-plugins/jdbc/src/main/scala/org/apache/linkis/manager/engineplugin/jdbc/executer/JDBCEngineConnExecutor.scala
+++ b/linkis-engineconn-plugins/engineconn-plugins/jdbc/src/main/scala/org/apache/linkis/manager/engineplugin/jdbc/executer/JDBCEngineConnExecutor.scala
@@ -27,12 +27,16 @@ import org.apache.linkis.manager.engineplugin.jdbc.ConnectionManager
 import org.apache.linkis.manager.engineplugin.jdbc.conf.JDBCConfiguration
 import org.apache.linkis.manager.label.entity.Label
 import org.apache.linkis.protocol.engine.JobProgressInfo
-import org.apache.linkis.rpc.Sender
+import org.apache.linkis.rpc.{RPCMapCache, Sender}
 import org.apache.linkis.scheduler.executer.{AliasOutputExecuteResponse, ErrorExecuteResponse, ExecuteResponse, SuccessExecuteResponse}
 import org.apache.linkis.storage.domain.{Column, DataType}
 import org.apache.linkis.storage.resultset.ResultSetFactory
 import org.apache.linkis.storage.resultset.table.{TableMetaData, TableRecord}
 import org.apache.commons.io.IOUtils
+import org.apache.linkis.common.conf.Configuration
+import org.apache.linkis.governance.common.protocol.conf.{RequestQueryEngineConfig, ResponseQueryConfig}
+import org.apache.linkis.manager.label.entity.engine.{EngineTypeLabel, UserCreatorLabel}
+import org.apache.linkis.protocol.CacheableProtocol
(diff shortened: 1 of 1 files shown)


linkis-engineconn-plugins/engineconn-plugins/jdbc/src/main/scala/com/webank/wedatasphere/linkis/manager/engineplugin/jdbc/executer/JDBCEngineConnExecutor.scala
```diff
@@ -177,3 +190,17 @@ class JDBCEngineConnExecutor(override val outputPrintLimit: Int, val id: Int) ex
...
 }
 
+
+object JDBCEngineConfig extends RPCMapCache[JobRequest, String, String](Configuration.CLOUD_CONSOLE_CONFIGURATION_SPRING_APPLICATION_NAME.getValue) {
+
+  override protected def createRequest(jobReq: JobRequest): CacheableProtocol = {
+    val engineTypeLabel = new EngineTypeLabel()
+    engineTypeLabel.setEngineType("jdbc")
+    engineTypeLabel.setVersion("4")
+    RequestQueryEngineTypeDefault(engineTypeLabel)
```
leeebai, line 200: This request can only get default configurations of jdbc engine, cannot get user configurations in governance station.
you can change to use RequestQueryEngineConfig. User and Creator labels can be found in EngineConnTask.

leeebai (changes requested):

linkis-engineconn-plugins/engineconn-plugins/jdbc/src/main/scala/com/webank/wedatasphere/linkis/manager/engineplugin/jdbc/executer/JDBCEngineConnExecutor.scala
FireFoxAhri: image
I've found that the jdbc url is passed as engineconn-conf, but when executing the code the conf is not used. The conf is parsed in EngineConnServer, is there any way to take the conf in TaskExecutionServiceImpl.

By the way. I've found another bug. If the jdbc url contains char like & ,eg /test?characterEncoding=UTF-8&useSSL=false, it will make the command execute in the backgroud and ignore chars behind &. Then the engineConn start failed.

leeebai: > image
> I've found that the jdbc url is passed as engineconn-conf, but when executing the code the conf is not used. The conf is parsed in EngineConnServer, is there any way to take the conf in TaskExecutionServiceImpl.
> 
> By the way. I've found another bug. If the jdbc url contains char like & ,eg /test?characterEncoding=UTF-8&useSSL=false, it will make the command execute in the backgroud and ignore chars behind &. Then the engineConn start failed.

Thank you very much for your careful research, I hope the following answers can help you.

1.The engine's default parameters are loaded during the engine startup, such as JDBC URL, UserName, Password, etc., this parameter is the default value in the configuration, rather than user-defined values, which are invalid. So we need to dynamically load the user's configuration during the task runtime, which is the real meaning of this Pull Request.

2.When the JDBC URL is dynamically obtained via the RPC, the & Symbol should not be lost. You can debug the JDBC engine and check the createDatasources method, which will make simple processing and storage of the URL, and then establish a JDBC connection directly.

leeebai (approved):
LGTM.

PR #943, opened 2021-08-07
Hive3.0 compiles and reports an error. Instead of the specific implementation class, WritableComparable accepts the return, and the compilation passes

zosimer:
image
The DateWritable class in hive3.1.2 has been abandoned and replaced by DateWritableV2. We can use the common parent class of DateWritableV2 and DateWritable to receive the return value, so that it can be compatible with previous and future versions of hive
case DATE: {
                WritableComparable dw = ((DateObjectInspector) oi).getPrimitiveWritableObject(o);
                binaryData = Base64.encodeBase64(String.valueOf(dw).getBytes());
                break;
            }
            case TIMESTAMP: {
                WritableComparable tw = ((TimestampObjectInspector) oi).getPrimitiveWritableObject(o);
                binaryData = Base64.encodeBase64(String.valueOf(tw).getBytes());
                break;
            }


peacewong (approved):
LGTM.

PR #956, opened 2021-08-16
fix a flink sql executing problem

### LINKED ISSUE #955: There is a flink sql executing problem
When I use Linkis flink engineconn to execute sql, I found there was a problem using specific grammar

yangzhiyue:
this pr is to fix a flink sql executing problem
close #955


peacewong:
The newly submitted PR has been merged

Issue #912, opened 2021-07-26
DataSource Web


Adamyuanyuan:
DataSource web module is the front-end display of the data source module, which is used by the user to manage and view the data source on the interface.

PR #985, opened 2021-09-01
datasource web

Adamyuanyuan:
In order to make it more convenient for users to manage data sources, we built a unified datasource module with the help of WeBank.These functions are included in the module：
DataSource web module is the front-end display of the data source module, which is used by the user to manage and view the data source on the interface. #912


Adamyuanyuan:
closed for linkis-1.1.0

Issue #726, opened 2021-05-11
Kafka data source


Adamyuanyuan:
**Describe the solution you'd like**
Added Kafka data source module, its role is to unify the management of Kafka data sources, Kafka is usually used for streaming computing applications;

Issue #725, opened 2021-05-11
DataSource Client


Adamyuanyuan:
**Describe the solution you'd like**
In order to make user more convenient for users to manage data sources, we built a unified DataSource module with the help of WeBank partners. The DataSource Client module contains the call logic of the Client side, so that users can operate data sources and obtain relevant metadata in the way of the Client side.

Issue #724, opened 2021-05-11
DataSource Metadata


Adamyuanyuan:
**Describe the solution you'd like**
In order to make user more convenient for users to manage data sources, we built a unified data source module with the help of WeBank partners, among which MetaManager module is used for docking metadata of various data sources in the big data platform, mainly including data source connection testing, metadata query and other functions;

Issue #723, opened 2021-05-11
Datasource manager 


Adamyuanyuan:
**Describe the solution you'd like**
In order to make user more convenient for users to manage data sources, we built a unified datasource module with the help of WeBank. DataSourceManager module is the core module of data sources, which is used to manage all kinds of data sources in the big data platform, mainly including the addition, deletion, modification and release of data sources.Version management and other functions;

PR #986, opened 2021-09-02
Dev 1.1.0-Linkis DataSource

Adamyuanyuan:
In order to make it more convenient for users to manage data sources, we built a unified datasource module with the help of WeBank.These functions are included in the module：

DataSourceManager module. DataSourceManager is the core module of data sources, which is used to manage all kinds of data sources in the big data platform, mainly including the addition, deletion, modification and release of data sources.Version management and other functions; Datasource manager  #723
MetaManager module is used for docking metadata of various data sources in the big data platform, mainly including data source connection testing, metadata query and other functions; DataSource Metadata #724
The DataSource Client module contains the call logic of the Client side, so that users can operate data sources and obtain relevant metadata in the way of the Client side; DataSource Client #725
Added Kafka data source module, its role is to unify the management of Kafka data sources, Kafka is usually used for streaming computing applications; Kafka data source #726


Adamyuanyuan:
closed for dev-1.1.0

Issue #1057, opened 2021-11-01
If an engineConn's lib dir have subdir then not all the jar could be up load to engineConnPublickDir


lordk911:
linkis 1.0.2 , Hdp3.1.4
In hdp env hive will default use atlas to save lineage , atlas's hive hook Impl jars is in a subdir, In this situation , the default behavior of linkis will got Exception

lordk911:
https://github.com/apache/incubator-linkis/pull/1045

PR #1058, opened 2021-11-02
fix issue 1057 : support subdir to be 'zip' and 'unzip'

lordk911:
fix issue [1057](https://github.com/apache/incubator-linkis/issues/1057#issue-1041204222) : support subdir to be 'zip' and 'unzip'

### DIFF
--- a/linkis-commons/linkis-common/src/main/java/org/apache/linkis/common/io/Fs.java
+++ b/linkis-commons/linkis-common/src/main/java/org/apache/linkis/common/io/Fs.java
@@ -52,4 +52,8 @@ public interface Fs extends Closeable {
 
     public abstract boolean renameTo(FsPath oldDest, FsPath newDest) throws IOException;
 
+    public abstract boolean mkdir(FsPath dest) throws IOException;
+
+    public abstract boolean mkdirs(FsPath dest) throws IOException;
+
 }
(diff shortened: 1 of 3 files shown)


linkis-commons/linkis-common/src/main/java/org/apache/linkis/common/io/Fs.java
```diff
@@ -52,4 +52,8 @@
 
     public abstract boolean renameTo(FsPath oldDest, FsPath newDest) throws IOException;
 
+    public abstract boolean mkdir(FsPath dest) throws IOException;
```
peacewong, line 55: Hello, the code implemented by the method is not uploaded

lordk911: interface Fs in module linkis-common have a direct implement abstract class FileSystem in module linkis-storage. The method add into Fs have already IN FileSystem  before this PR

lordk911: I have remove the the abstract method mkdir\mkdirs from abstract class FileSystem and lift them to interface

peacewong (approved):
LGTM.

PR #1059, opened 2021-11-02
Support kerberos proxy user in Flink engine

saLeox:
Support kerberos proxy user in Flink engine.

### Brief change log
- If kerberos proxy user is enabled, will add current user as proxy user, and append it into engine jar start command as new environment variables, so that the method UserGroupInformation.getCurrentUser().getUserName() can fetch.

### Verifying this change
(example:)  
- Make sure your cluster has integrated with kerberos;
- Remove the remark of kerberos setting, switch the wds.linkis.keytab.host.enabled and wds.linkis.keytab.proxyuser.enable onto true;
- Set security.kerberos.login.use-ticket-cache: true in flink-conf.yaml, but don't specify the keytab.
- Use Linkis API to submit Flink task, then the USER in Yarn UI will be the current user in Linkis.

### Documentation


peacewong (approved):
LGTM.

PR #1093, opened 2021-11-16
Dev 1.0.3 bugfix passauthuri add judgement of whether it is empty

lunescode:
bugfix:
if the pass auth url set value “”, the method ".getRequestURI.startsWith" will always return true, it is not the result that we wanted, it can be a security risk.

### Brief change log
(for example:)
-  Add  judgement of whether it is empty

### Verifying this change
-  1. set wds.linkis.gateway.conf.url.pass.auth=""  in linkis-mg-gateway.properties, the restart gateway.
-  2. set wds.linkis.server.user.restful.uri.pass.auth="" in linkis-ps-publicservice.properties, the restart publicservice.
-  3. curl http://${gateway}/api/rest_j/v1/configuration/getCategory, if return "need login in" will be right.

- Dependencies (does it add or upgrade a dependency): ( no)
- Anything that affects deployment: ( no )

### Documentation
- Does this pull request introduce a new feature? ( no)
- If yes, how is the feature documented? (not documented)


peacewong (approved):
LGTM.

Issue #1089, opened 2021-11-15
deploy engine without workDir directory  write permission


casionone:
**Describe the bug**
execute engineConnExec.sh with permisson denied.
cause by workDir directory withot write permission

lunescode:
if need change permision , you can adjust the top directory permission.

PR #1094, opened 2021-11-16
Revert "[ISSUE#1089 ]bugfix:local file permission"

lunescode:
revert #1089 
it will be affect other function!

1. engineConnExec.sh default --rw-r--r-- permission, it is default  executed  by "sudo su - user -c "sh /app/linkis/tmp/ods/workDir/28225ed5-0e8b-48c5-8b90-07676b343f6e/engineConnExec.sh"" ,so it just need execute permission.
2.  workDir write permmision. you can change workDir's parent directory permission. can set 775, delte workDir,  and it will be created ,the permission will be as  same as parent.


peacewong (approved):
LGTM.

PR #1121, opened 2021-11-29
Get protocol from current request, remove hard code of 'http'

saLeox:
Given that there is an another outer Nginx, which enables HTTPS and forward request to Linkis Nginx, when Linkis front end calls the backend API, using the address from the ${window.location.host}, the protocol also need to align with the current request protocol, which can be gotten from ${window.location.protocol}. Otherwise, there will be a CORS error on broswer side as using https to request front end resources while using http to access backend API.  

### Brief change log
- In front end file api.js, change the protocol from 'http' to current request protocol.

### Verifying this change
(Please pick either of the following options)  
- Setup an outer Nginx, that enabled HTTPS, and forward requests to the Linkis Nginx, that just support HTTP. See whether request can reach the backend from broswer.

### Documentation

### DIFF
--- a/web/src/common/service/api.js
+++ b/web/src/common/service/api.js
@@ -50,7 +50,7 @@ let cutReq = (config) => {
 };
 
 const instance = axios.create({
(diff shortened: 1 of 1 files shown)


peacewong (approved):
LGTM.

PR #1149, opened 2021-12-03
fix: After the job task  completed , the front cannot obtain the prog…

zhaoyun006:

1. fix: After the job task  completed , the front cannot obtain the progess info data
2. /entrance/{id}/progress code map optimapzation


peacewong (approved):
LGTM.

Issue #1158, opened 2021-12-06
[Bug] NPE of api '/linkisManager/rm/engines'


casionone:
**Describe the bug**
When get engine nodes include node EMnode was empty will cause NullPointerException

**Expected behavior**
When it is empty, record map does not store this attribute

**Screenshots**
![image](https://user-images.githubusercontent.com/7869972/144820216-6febf552-694a-49be-b137-2bb136d87150.png)

PR #1168, opened 2021-12-07
fix NPE for /linkisManager/rm/engines api

legendtkl:
for issue https://github.com/apache/incubator-linkis/issues/1158


peacewong (approved):
LGTM.

Issue #1157, opened 2021-12-06
[Bug]Shell engine  will not append shell-stdout into persisted-log


casionone:
**Describe the bug**
Shell engine will not append shell-stdout to the persistent log when executing

**Expected behavior**
When the Shell engine is executed, shell-stdout needs to be appended to the persistent log

**Screenshots**

![企业微信截图_16387793782508](https://user-images.githubusercontent.com/7869972/144819108-d32a1434-7921-4d56-87fa-5ff1e606ca03.png)

legendtkl:
Hi, @casionone @peacewong, would you help review this pr?

Thanks, kelu.

peacewong:
> Hi, @casionone @peacewong, would you help review this pr?
> 
> Thanks, kelu.

ok thanks

PR #1172, opened 2021-12-08
Append shell stdout to persistent log

legendtkl:
fix for https://github.com/apache/incubator-linkis/issues/1157


linkis-engineconn-plugins/engineconn-plugins/shell/src/main/scala/org/apache/linkis/manager/engineplugin/shell/executor/ShellEngineConnExecutor.scala
```diff
@@ -77,6 +77,7 @@ class ShellEngineConnExecutor(id: Int) extends ComputationExecutor with Logging
       }) {
         info(line)
         engineExecutionContext.appendTextResultSet(line)
+        engineExecutionContext.appendStdout(line)
```
peacewong, line 80: In order to reduce the number of messages, it is recommended to modify it to push by log.`LogHelper.logCache.cacheLog(line)`

peacewong (approved):
LGTM.

Issue #1143, opened 2021-12-02
[Bug]Cannot be processed when the submitted task script is empty code


casionone:
**Describe the bug**
Cannot be processed when the submitted task script is empty code.
This scenario should return a prompt message that the script cannot be empty and prohibit submission or end the task directly.
**Expected behavior**
When this method(org.apache.linkis.engineconn.computation.executor.execute.ComputationExecutor#toExecuteTask ) is executed, it should be judged whether the executed code is empty.

legendtkl:
hi, @peacewong , would you help review this ?

thanks. kelu.

PR #1173, opened 2021-12-08
check the task code and make sure it's not empty

legendtkl:
fix https://github.com/apache/incubator-linkis/issues/1143


linkis-computation-governance/linkis-engineconn/linkis-computation-engineconn/src/main/scala/org/apache/linkis/engineconn/computation/executor/execute/ComputationExecutor.scala
```diff
@@ -145,6 +145,11 @@ abstract class ComputationExecutor(val outputPrintLimit: Int = 1000) extends Acc
 
 
   def toExecuteTask(engineConnTask: EngineConnTask, internalExecute: Boolean = false): ExecuteResponse = {
+    if (engineConnTask.getCode.trim.isEmpty) {
+      error(s"The code must not be empty!")
+      return ErrorExecuteResponse("The code must not be empty!", null)
```
peacewong, line 150: Hello, here to consider the case of submitting comment code

peacewong (approved):
LGTM.

PR #1186, opened 2021-12-10
Fix the issue that the tasks queued by orchestrator cannot be killed

### LINKED ISSUE #1182: Tasks in the orchestrator queue cannot be killed
**Describe the bug**
The tasks in the queue of the orchestrator cannot be killed, and the kill will fail and the task will continue to run.

**To Reproduce**
1. Submit multiple tasks 
2. kill queued task

**Expected behavior**
task be killed.

peacewong:
Fix the issue that the tasks queued by orchestrator cannot be killed close #1182

### Brief change log
- update ExecutionImpl to support killed queued task;


leeebai (approved):
LGTM.

PR #1187, opened 2021-12-10
The python engine does not use pandas. Delete this module

Zosimer:
The python engine does not use pandas. if you not intsall  pandas in you ststem, they will accident   errCode: 60003 ,desc: Python process is not alive，so Delete this module


peacewong (approved):
LGTM.

PR #1190, opened 2021-12-11
Solve the problem that the new user does not have a workspace directory

Zosimer:
When an LDAP user adds a new user, the corresponding user does not add a workspace directory. As a result, the new user cannot obtain his own workspace directory after logging in. I found that if there is no corresponding directory, he will create his own workspace directory for the corresponding user


peacewong:
Hello, user initialization is recommended to be placed in a separate process

PR #1191, opened 2021-12-11
Set the spark related memory to 1g and the concurrency to 1core

Zosimer:
In order to reduce the probability of failed startup of linkis caused by insufficient resources, improve the success rate of linkis installation, and minimize the resource settings related to spark


peacewong (approved):
LGTM.

PR #1199, opened 2021-12-14
Fix issue SqlCodeParser has a bug that splits the escaped ";" into multiple SQL

### LINKED ISSUE #1197: [Bug][Entrance]SqlCodeParser has a bug that splits the escaped ";" into multiple SQL
**Describe the bug**
There is a bug in SqlCodeParser that splits the escaped ";" into multiple sql, resulting in a syntax error in sql runtime.

**To Reproduce**
1. submit sql task ```select  "\;"```
2. task running
3. See error
企业微信截图_16393945939357

peacewong:
 close #1197 

### Brief change log
- Optimize SqlCodeParser


Alexkun (approved):
LGTM.

Issue #1208, opened 2021-12-15
Help methods to increase the proportion of resources obtained


peacewong:
**Describe the solution you'd like**
Get the proportion of left resources, and return the smallest CPU, memory, and instance

**Describe alternatives you've considered**
Method to increase the proportion of resources obtained，A clear and concise description of any alternative solutions or features you've considered.

Issue #1207, opened 2021-12-15
The queue name in YarnResource should not be used as calculation logic


peacewong:
**Describe the bug**
There is an incorrect usage of using the queue name as a comparison in YarnResource
```
override def notLess(r: Resource): Boolean =
    queueName >= r.queueName && queueMemory >= r.queueMemory && queueCores >= r.queueCores && queueInstances >= r.queueInstances

```

Issue #1202, opened 2021-12-14
It is recommended to optimize the messageScheduler as a direct service call


peacewong:
**Is your feature request related to a problem? Please describe.**
There are some internal servive calls in the manager that use the messagescheduler, and there are performance problems during high concurrency.

**Describe the solution you'd like**
Optimized to call the service layer directly

Issue #1200, opened 2021-12-14
It is recommended to add update and startup time attributes to the node object


peacewong:
**Is your feature request related to a problem? Please describe.**
Node should record the corresponding time when it is created and modified

**Describe the solution you'd like**
add update and startup time attributes to the node object

PR #1209, opened 2021-12-15
Optimize multiple features of manager common

peacewong:
1. add update and startup time attributes to the node object #1200 
2. Remove the judgment logic of the queue name #1207 
3. Method to increase the proportion of resources obtained #1208 
4. Add em registration and cleanup requests #1202 

### Brief change log
- Optimize multiple class files of manager common module;


Alexkun (approved):
LGTM.

PR #1213, opened 2021-12-16
Optimized support for relationship between the long-lived label and the node not be deleted

### LINKED ISSUE #1212: It is recommended that the relationship between the long-lived label and the node not be deleted
**Describe the bug**
The relationship between tenant/route label and node should not be deleted

**Expected behavior**
When removing the node label, you should keep the long-lived label.

peacewong:
Optimized support for relationship between the long-lived label and the node not be deleted
close #1212 

### Brief change log
- Optimize and adjust the labelmanager code


leeebai (approved):
LGTM.

PR #1232, opened 2021-12-22
Use current user to read result file, and abstract the file permission in filesystem

saLeox:
For now, the file permission of the generated result set in HDFS or local file system is not specified in one place globally, and followed the default value (744), and JVM user is used to read those result file instead of current login user.
It causes two issues:

1) The result set is transparent to all hdfs users;
2) JVM user cannot read current login user's result set when the permission of file is changed to more strict value(like 740).

### Brief change log
- Specify the permission value in FileSystem (740 for file and 750 for folder) , and apply it in in local file system and hdfs;
- Use current login user to read result file instead of JVM user.

### Verifying this change
(Please pick either of the following options)  
- Submit any type of task normally, and check whether of permission value of results set or log in the hdfs is 740, and it's expected to be available to current login user.

### Documentation


peacewong (approved):
LGTM.

PR #1234, opened 2021-12-24
Fix this issue that  memory leak problem of SSO login

### LINKED ISSUE #1138: [Bug]The memory leak problem of SSO login
**Describe the bug**
There is a memory leak in the Map of the session record data of the Gateway's SSO login. The session is not cleaned up, resulting in an Out Of Memory Exception of the service jvm

peacewong:
Fix this issue that  memory leak problem of SSO login close #1138 

### Brief change log
- update map to ConcurrentHashMap.


leeebai (approved):
LGTM.

Issue #1241, opened 2021-12-28
Invalid default value for 'created_time'


casionone:
**Describe the bug**
ddl sql excute with error [Invalid default value for 'created_time']
mysql version 5.7.31

PR #1243, opened 2021-12-28
fix ddl datetime(3) default value

peacewong:
fix ddl: datetime(3) default value
Related issues: #1241


liuyou2 (approved):
LGTM

PR #1247, opened 2021-12-28
git ignore target for parent pom

legendtkl:
ignore the target dir generated by 'mvn -N install' in parent pom.


peacewong (approved):
LGTM.

Issue #1262, opened 2021-12-31
[Bug] when uploading large bml resource files  causes oom exception


casionone:
**Describe the bug**

when uploading large bml resource files  causes oom exception
```
ResponseBody is {"timestamp":1640865932274,"status":500,"error":"Internal ServerError","message":"","path":"/api/rest_j/v1/bml/upload"}
Caused by: java.lang.OutOfMemoryError: Java heap space
at java.util.Arrays.copyOf(Arrays.java:3236) ~[?:1.8.0_261] 
at org.eclipse.jetty.util.MultiPartInputStreamParser$MultiPart.write(MultiPartInputStreamParser.java:173) ~[jetty-util-9.4.30.v20200611.jar:9.4.30.v20200611]
```

PR #1263, opened 2021-12-31
Fix:oom exception when upload file #1262

casionone:
Fix the bug: when uploading large bml resource files causes oom exception
Related issues: #1262 

### Brief change log
New configuration spring.spring.servlet.multipart.file-size-threshold=50MB.
The default value is 0, Jetty and tomcat containers handle the 0 value of this parameter differently
Tomcat will write to disk immediately
Jetty will never write to disk
See for details: https://github.com/spring-projects/spring-boot/issues/9073

- Anything that affects deployment: (no )


peacewong (approved):
LGTM.

PR #1264, opened 2022-01-01
Pyspark  support CDH and HDP versions

Zosimer:
Because pyspark obtains the spark version number through ‘spark submit -- version’, and converts the version number into an integer. For example, spark-2.4.1 has a version number of 241, but the spark version of CDH is 2.4.1-cdh6 3.2 this type cannot be converted to integer. The spark version of HDP is 2.3.2.3.1.4.0-315. After conversion, it is 2.3.2.3.1.4.0-315, and then an error will be reported。
Therefore, I withdraw the top five digits of the version number, such as 2.4.1-cdh6 3.2, take 2.4.1 as the version number, which is similar to HDP and can meet the requirements。

9a9a2142e4266c274bb1f53c18ce691

d9482a1ee0081564bf6a8a424442173

### DIFF
--- a/linkis-engineconn-plugins/engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/executor/SparkPythonExecutor.scala
+++ b/linkis-engineconn-plugins/engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/executor/SparkPythonExecutor.scala
@@ -161,7 +161,7 @@ class SparkPythonExecutor(val sparkEngineSession: SparkEngineSession, val id: In
     val cmd = CommandLine.parse(pythonExec)
     cmd.addArgument(createFakeShell(pythonScriptPath).getAbsolutePath, false)
     cmd.addArgument(port.toString, false)
-    cmd.addArgument(EngineUtils.sparkSubmitVersion().replaceAll("\\.", ""), false)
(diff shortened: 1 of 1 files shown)


linkis-engineconn-plugins/engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/executor/SparkPythonExecutor.scala
```diff
@@ -161,7 +161,7 @@ class SparkPythonExecutor(val sparkEngineSession: SparkEngineSession, val id: In
     val cmd = CommandLine.parse(pythonExec)
     cmd.addArgument(createFakeShell(pythonScriptPath).getAbsolutePath, false)
     cmd.addArgument(port.toString, false)
-    cmd.addArgument(EngineUtils.sparkSubmitVersion().replaceAll("\\.", ""), false)
```
peacewong, line 164: This bug can be modified by referring to the [ 0.11.0 method](https://github.com/apache/incubator-linkis/commit/0486a9abf3c2c2e1261195e41e758f5b9f0f9c78#diff-597ca2ca884fe266f156082c9c6a316435116a41c9ae56e8cbc9a56eea23fbee)

PR #1270, opened 2022-01-04
streamExecEnv is not initialized, resulting in null pointer exception

Zosimer:
in class ClusterDescriptorAdapter ，this line on
configuration = (Configuration) method.invoke(executionContext.getStreamExecutionEnvironment()) ，this code ’executionContext.getStreamExecutionEnvironment()‘ return null，beacause streamExecEnv is not initialized, resulting in null pointer exception
image


wushengyeyouya (approved):
LGTM

PR #1288, opened 2022-01-10
Fix the bug when ticketId starts with number, the engineConnId cannot be parsed successful.

wushengyeyouya:
Fix the bug when ticketId starts with number, the engineConnId cannot be parsed successful

### Brief change log
- Fix the bug when ticketId starts with number, the engineConnId cannot be parsed successful.


peacewong (approved):
LGTM.

Issue #1285, opened 2022-01-10
[Bug] Make the LICENSE file different for the source release and the binary release


casionone:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-commons

### What happened + What you expected to happen

Only licenses from software included in the source release need to be mentioned in LICENSE and NOTICE not all dependancies. This may make the LICENSE file different for the source release and the binary release.
Detail info: https://lists.apache.org/thread/pcopkzcfzq793rbfmx9szm857oynrn23 

### Relevent platform

none

### Reproduction script

none

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1295, opened 2022-01-11
Remove old web licenses file amd fix cpoy bug when build web

casionone:
Remove old web licenses file amd fix cpoy bug when build web
web licenses files at licenses-binary-ui/*
- #1285


peacewong (approved):
LGTM.

PR #1297, opened 2022-01-12
Modify upload file max size limit to 500M

casionone:
The lib package of some engine plugin  is currently larger than 200M, and adjust  spring.servlet.multipart.max-file-size to 500M


peacewong (approved):
LGTM.

PR #1304, opened 2022-01-14
Fix some bug of Flink EC and optimize the computation OnceJob.

wushengyeyouya:
Fix some bug of Flink EC and optimize the computation OnceJob.

### Brief change log
- Fix the bug when ticketId starts with number, the engineConnId cannot be parsed successful.
- Fix the bug that flink sql cannot support multi-sql and optimize some code formats.
- Fix the bug that when use set, rest, drop grammar, the checkpoint will invalid.
- Fix the bug that when operator is failed and the job is not instanced , will cause the flink yarn application stop failed when engineconn exits .
- Add the necessary dependency of minlog in flink EngineConn.


peacewong (approved):
LGTM.

PR #1306, opened 2022-01-14
modify ’wds.linkis.entrance.skip.auth‘ default false change to true

Zosimer:
when I use python moudle  os and sys ,then report error 'can not use sys module' and ’can not use os module‘，so we modified  ’wds.linkis.entrance.skip.auth‘ default false change to true


peacewong (approved):
LGTM.

PR #1436, opened 2022-02-11
Solve the problem of java code format

peacewong:
Solve the problem of java code format

### Brief change log
- use spotless plugin to format java code;


hongxiaa:
邮件已经收到，谢谢

PR #1459, opened 2022-02-15
Set linkis version from 1.0.3 to 1.1.0

casionone:
Set linkis version from 1.0.3 to 1.1.0

Use the mvn tool to upgrade the version uniformly Execution command example:
```
$ mvn versions:set -DnewVersion=1.1.0
#Modify the configuration in the parent pom.xml
<linkis.version>1.1.0</linkis.version>
```


peacewong (approved):
LGTM.

Issue #1477, opened 2022-02-16
[Feature] Adjust the package name of the message


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Now the package name of message in `org.apache.linkis.rpc.serializer` does not match the feature relationship, it is recommended to change it to `org.apache.linkis.rpc.message`.

### solutions

Modify the package name of message to org.apache.linkis.rpc.message.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1478, opened 2022-02-16
Adjust the package name of the message

peacewong:
Adjust the package name of the message #1477 

### Brief change log
- Modify the package name of message to org.apache.linkis.rpc.message.
- Optimize ecp module package import .


Alexkun (approved):
LGTM.

PR #1510, opened 2022-02-18
support all type url

det101:
datasource url support all types

### Brief change log
The interface of the database configuration data source is supported, and the corresponding request address is
/xxx ->   http://{window.location.origin}/xxx
/xxx ->   http://{window.location.href}/xxx
http://xxx -> http://xxx
JSONString ->  Object


peacewong (approved):
LGTM.

Issue #1559, opened 2022-02-25
[Feature] Optimize the one-click deployment install.sh script


casionone:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Some optimizations such as:
- Before performing database initialization, print the information of the database so that personnel can verify it again
- Perform a connection test of the database before deployment
- Add necessary deployment log information, such as logs of execution steps/create directories, etc.
- Necessary key information key color identification
- Add basic environment check for spark/hadoop/hive
- deploy-conf: migrate HIVE_META configuration to db.sh

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1561, opened 2022-02-25
optimize install.sh script

casionone:
optimize install.sh script
related issues: #1559


peacewong (approved):
LGTM

Issue #1574, opened 2022-02-28
[Feature][Mybatis]Added DataSourceUtils tool class for creating datasource


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Now datasource is created in both mybatis and metadat modules and the code is repeated. It is recommended to optimize the code and implement common logic.
现在datasource创建在mybatis和metadat模块都存在而且代码重复，建议优化代码，实现公共创建datasource的逻辑

### Description

Added DataSourceUtils tool class for creating datasource

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

Issue #1537, opened 2022-02-21
[Feature] Gateway supports adding login user suffix after user login


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

There are some scenarios where users need to use username + suffix after logging in, such as adding mailboxes or adding group information, etc.

### solutions

Gateway supports adding login user suffix after user login

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1575, opened 2022-02-28
Added DataSourceUtils tool class for creating datasource

peacewong:
Added DataSourceUtils tool class for creating datasource #1574 
Gateway supports adding login user suffix after user login #1537

### Brief change log
- Added DataSourceUtils tool class for creating datasource
- Gateway supports adding login user suffix after user login
- Add label information of openLookeng EC


liuyou2 (approved):
LGTM

Issue #1598, opened 2022-03-03
[Bug] The version of commons-compress is too low, resulting in an error when the result set is downloaded as excel


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-commons

### What happened + What you expected to happen

1. Download the result set as excel
2. The background error is as follows
`Throw error java.lang.NoClassDefFoundError: org/apache/commons/compress/utils/InputStreamStatistics`

3. commons-compress is caused by the low version introduced by hadoop-common

### Relevent platform

all

### Reproduction script

see happened 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1600, opened 2022-03-03
Fix this isse caused by the low version of commons-compress

peacewong:
#1598 

### Brief change log
- exclude commons-compress in hadoop-common


Adamyuanyuan (approved):
thanks

Adamyuanyuan:
thanks，merged

Issue #1601, opened 2022-03-03
[Bug] client unsupport runtimeMap param


LongJGun:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-client

### What happened + What you expected to happen

The CLI client does not support runtimeMap. As a result, runtime parameters cannot be configured
The JDBC engine is having trouble dynamically configuring connection information

cli客户端不支持runtimeMap参数，导致runtime参数不能配置
jdbc引擎要动态配置连接信息遇到困难

### Relevent platform

centos7

### Reproduction script

-runtimeMap jdbc.url=xxxx -runtimeMap jdbc.username=xxxx 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1603, opened 2022-03-04
fix client unsupport -runtimeMap param #1601

ganlangjie:
fix client unsupport **-runtimeMap** param
[https://github.com/apache/incubator-linkis/issues/1601](url)


peacewong:
> fix client unsupport **-runtimeMap** param [https://github.com/apache/incubator-linkis/issues/1601](url)

Thanks!

Fuu3214 (changes requested):
Thx for the commit, but I’m afraid this won’t work. You need to pass input runtimeMap params to LinkisJob via VarAccess. You can refer to code in LinkisJobBuilder. 
By the way, we have already added -runtimeMap option in 1.1.1 version. And it will be released very soon. So you can also put this problem aside and wait for us to handle it.

Shang

ganlangjie:
OK

peacewong (approved):
LGTM.

Issue #1609, opened 2022-03-04
[Bug] jdbc engine postgresql unsupport sql "show databases" 


ganlangjie:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-engineConnplugin

### What happened + What you expected to happen

![image](https://user-images.githubusercontent.com/22737636/156735227-3d9222ee-fb93-4c2b-b574-147af5066d22.png)

DBCP针对不同数据库的validationQuery写法   
postgresql 不支持语法 show databases   导致 执行失败

### Relevent platform

centos 7

### Reproduction script

sh ./bin/linkis-cli -engineType jdbc-4 -codeType jdbc -code "SELECT * FROM person"  -submitUser root -proxyUser root -runtimeMap jdbc.url=jdbc:postgresql://192.168.1.5:5432/postgres?currentSchema=public -runtimeMap jdbc.username=gpadmin -runtimeMap jdbc.password=gpadmin  -runtimeMap jdbc.auth.type=USERNAME

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1610, opened 2022-03-04
fix jdbc engine postgresql unsupport sql "show databases;" lead to exec fail

ganlangjie:
fix jdbc engine postgresql unsupport sql "show databases;" lead to exec fail #1609


peacewong (approved):
LGTM.

PR #1635, opened 2022-03-08
[Feature] Use regex to match against gateway URLs, if matched, let them pass

silly-carbon:
This PR is based on https://github.com/apache/incubator-linkis/pull/556

This BUG is fixed in this PR. With REGEX Support.

### DIFF
--- a/linkis-spring-cloud-services/linkis-service-gateway/linkis-gateway-core/src/main/scala/org/apache/linkis/gateway/config/GatewayConfiguration.scala
+++ b/linkis-spring-cloud-services/linkis-service-gateway/linkis-gateway-core/src/main/scala/org/apache/linkis/gateway/config/GatewayConfiguration.scala
@@ -63,4 +63,7 @@ object GatewayConfiguration {
 
   val DSS_QUERY_WORKSPACE_SERVICE_NAME = CommonVars("wds.dss.query.workspace.service", "dss-framework-project-server")
   val USER_WORKSPACE_REFLESH_TIME  = CommonVars("wds.linkis.user.workspace.reflesh.time", 10)
+
+  // Use regex to match against URLs, if matched, let them pass anyway(even if not currently logged in)
+  val GATEWAY_NO_AUTH_URL_REGEX = CommonVars("wds.linkis.gateway.no.auth.url.regex", ".*visualis.*share.*")
 }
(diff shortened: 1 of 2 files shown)


silly-carbon:
@peacewong Please see this PR

peacewong:
> @peacewong Please see this PR

ok. Thanks

linkis-spring-cloud-services/linkis-service-gateway/linkis-gateway-core/src/main/scala/org/apache/linkis/gateway/config/GatewayConfiguration.scala
```diff
@@ -63,4 +63,7 @@ object GatewayConfiguration {
 
   val DSS_QUERY_WORKSPACE_SERVICE_NAME = CommonVars("wds.dss.query.workspace.service", "dss-framework-project-server")
   val USER_WORKSPACE_REFLESH_TIME  = CommonVars("wds.linkis.user.workspace.reflesh.time", 10)
+
+  // Use regex to match against URLs, if matched, let them pass anyway(even if not currently logged in)
+  val GATEWAY_NO_AUTH_URL_REGEX = CommonVars("wds.linkis.gateway.no.auth.url.regex", ".*visualis.*share.*")
```
peacewong, line 68: Hello, is it better to consider configuring multiple regulars?

peacewong:
Hello, can you help submit it to the dev-1.1.0-datasource branch, which will be released soon

linkis-spring-cloud-services/linkis-service-gateway/linkis-gateway-core/src/main/scala/org/apache/linkis/gateway/config/GatewayConfiguration.scala
silly-carbon: Not sure if I understand your meaning. Because REGEX already supported the OR relation(So you could match  against multiple URLs with just one pattern), which eliminated the need for "configuring multiple regulars"

Could you provide some examples?

peacewong: > Not sure if I understand your meaning. Because REGEX already supported the OR relation(So you could match against multiple URLs with just one pattern), which eliminated the need for "configuring multiple regulars"
> 
> Could you provide some examples?

Yes, that's what it means, you can provide an example in the comments

silly-carbon: With "()" and "|" you can express the "OR" relation in REGEX, for instance:

(.*visualis.*|.*dss.*) will MATCH URLs containing visualis OR  dss, the following URLS with be matched:
1. http://linkis.org/jd/visualis/vis
2. http://linkis.org/jd/dss/vis

Other URLs will NOT be matched.

PR #1636, opened 2022-03-08
[Feature] Use regex to match against gateway URLs, if matched, let them pass

silly-carbon:
This PR is based on https://github.com/apache/incubator-linkis/pull/1635

This BUG is fixed in this PR. With REGEX Support.

With "()" and "|" you can express the "OR" relation in REGEX, for instance:

(.visualis.|.dss.) will MATCH URLs containing visualis OR dss, the following URLS with be matched:

1. http://linkis.org/jd/visualis/vis
2. http://linkis.org/jd/dss/vis

Other URLs will NOT be matched.


silly-carbon:
@peacewong See this.

peacewong:
> @peacewong See this.

ok,Thanks

peacewong (approved):
LGTM.

Issue #1726, opened 2022-03-14
[Feature] /api/rest_j/v1/udf/all API suggested to optimize performance


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

There is a performance problem with the /api/rest_j/v1/udf/all API interface, and sometimes the response takes more than 10S

### Description

suggested to optimize performance

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1728, opened 2022-03-14
Optimize the performance of API /api/rest_j/v1/udf/all

peacewong:
#1726 

### Brief change log
- Optimize the performance of API /api/rest_j/v1/udf/all


Alexkun (approved):
LGTM.

Issue #1730, opened 2022-03-15
The python engine's showDF built-in function field lines are reversed


huiyuanjjjjuice:
When calling the python engine's showDF built-in function, the result set table field rows are reversed

PR #1731, opened 2022-03-15
Fix the problem that the result set field row of the python engine's showDF function is reversed

huiyuanjjjjuice:
fix  issue #1730


peacewong (approved):
LGTM.

Issue #1737, opened 2022-03-16
[Bug] [dev-1.1.1]There is a memory leak problem in the logout api


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-service-gateway

### What happened + What you expected to happen

Because the logout interface has a wrong loop call problem, affecting branch 1.1.1
```
SSOUtils
def removeLoginUserByAddCookie(addEmptyCookie: Cookie => Unit): Unit = {
    val cookie = new Cookie(USER_TICKET_ID_STRING, null)
    cookie.setMaxAge(0)
    cookie.setPath("/")
    if(sslEnable) cookie.setSecure(true)
    addEmptyCookie(cookie)
    removeLoginUserByAddCookie(addEmptyCookie)
  }
```

### Relevent platform

all

### Reproduction script

request logout

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1744, opened 2022-03-17
Dev 1.1.1 fix gateway oom

peacewong:
#1737 

### Brief change log
- conf openlookeng default multi.user 
- add presto/datax/sqoop version 
- update hive unique column default true 
- Fix this issue that memory leak problem #1737


casionone (approved):
LGTM

Issue #1638, opened 2022-03-09
[Bug][Linkis-PS-Metadatamanager]MetaMethodInvokeException:wrong number of arguments


casionone:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-public-enhancements

### What happened + What you expected to happen

Fail to invoke method: [getPartitions] in meta service instance: [org.apache.linkis.metadatamanager.service.MysqlMetaService]
with wrong number of arguments
```
 public MetaPartitionInfo getPartitions(
            String operator,
            Map<String, Object> params,
            String database,
            String table,
            boolean traverse) {
        return this.getConnAndRun(
                operator, params, conn -> this.queryPartitions(conn, database, table, traverse));
    }
```
The parameter traverse was not passed through the client entity class MetadataGetPartitionsAction

### Relevent platform

None

### Reproduction script

```    
new LinkisMetaDataRemoteClient(clientConfig)
.getPartitions(MetadataGetPartitionsAction.builder()
.setUser("xxx")
.setDataSourceId(5)
.setDatabase("linkis")
.setSystem("system")
.setTable("linkis_datasource").build()).
getProps
```

### Anything else

None 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #1778, opened 2022-03-22
fix:DWSHttpMessageResult cannot distinguish different methods(GET/DELETE) of the same RequestMapping request

casionone:
fix: DWSHttpMessageResult cannot distinguish different methods(GET/DELETE) of the same RequestMapping request 
fix: Datasource getPartitionsByDsId:add traverse parameter #1638


peacewong (approved):
LGTM.

Issue #1779, opened 2022-03-23
[Bug] [ECM] ECMNodeHeartbeatMonitor monitoring ECM should consider heartbeat reporting time


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-engineConnManager

### What happened + What you expected to happen

NodeHeartbeatMonitor monitoring ECM should consider the heartbeat reporting time. If you try to judge whether the eureka list exists, it may be wrongly judged due to the performance problem of eureka.
NodeHeartbeatMonitor监控ECM应该考虑心跳上报时间，如果试试判断eureka列表是否存在，可能因为eureka的性能问题导致错误判断.

### Relevent platform

all

### Reproduction script

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

Issue #1765, opened 2022-03-21
[Bug] [EC]EC tryShutdown will have blocking problems


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-engineConnplugin

### What happened + What you expected to happen

EC tryShutdown has a blocking problem. When the task is running and triggers shutdown, it will block. The jstack log is as follows:
EC tryShutdown 会存在阻塞问题，当又任务正在运行触发shutdown时会阻塞，jstack日志如下：
```
SpringContextShutdownHook" #28 prio=5 os_prio=0 tid=0x00007fc16402c800 nid=0x789e waiting for monitor entry [0x00007fc1ed5e0000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at org.apache.linkis.engineconn.executor.entity.SensibleExecutor$class.transition(SensibleExecutor.scala:36)
        - waiting to lock <0x00000000c19bf838> (a org.apache.linkis.manager.engineplugin.python.executor.PythonEngineConnExecutor)
        at org.apache.linkis.engineconn.acessible.executor.entity.AccessibleExecutor.transition(AccessibleExecutor.scala:27)
        at org.apache.linkis.engineconn.computation.executor.execute.ComputationExecutor.tryShutdown(ComputationExecutor.scala:97)
        at org.apache.linkis.engineconn.acessible.executor.service.DefaultAccessibleService$$anonfun$executorShutDownHook$1.apply$mcZ$sp(DefaultAccessibleService.scala:83)
        at org.apache.linkis.engineconn.acessible.executor.service.DefaultAccessibleService$$anonfun$executorShutDownHook$1.apply(DefaultAccessibleService.scala:83)
        at org.apache.linkis.engineconn.acessible.executor.service.DefaultAccessibleService$$anonfun$executorShutDownHook$1.apply(DefaultAccessibleService.scala:83)
        at org.apache.linkis.common.utils.Utils$.tryCatch(Utils.scala:40)
        at org.apache.linkis.common.utils.Utils$.tryAndWarn(Utils.scala:69)
        at org.apache.linkis.engineconn.acessible.executor.service.DefaultAccessibleService.executorShutDownHook(DefaultAccessibleService.scala:82)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #1780, opened 2022-03-23
Fix issue that EC tryShutdown will have blocking problems

peacewong:
#1779
#1765

### Brief change log
- ECMNodeHeartbeatMonitor monitoring ECM should consider heartbeat reporting time #1779
- update HIVE_RESULTSET_USE_UNIQUE_COLUMN_NAMES default value 
- Fix issue that EC tryShutdown will have blocking problems #1765


Alexkun (approved):
LGTM.

Issue #1801, opened 2022-03-25
[Bug] [ECM]ECM starts ec If the status is Failed, it should kill ec process


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-engineConnManager

### What happened + What you expected to happen

ECM starts ec If the status is Failed, it should kill the ec process. For example, timeout due to insufficient queue resources
ECM 启动ec如果状态为Failed，应该kill ec process。比如因为队列资源不足超时

### Relevent platform

all

### Reproduction script

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1806, opened 2022-03-28
Fix issue EC startup failure was not properly cleaned up

peacewong:
ECM starts ec If the status is Failed, it should kill ec process #1801


Alexkun (approved):
LGTM.

Issue #1813, opened 2022-03-28
[Bug] [AM]EngineRestfulApi createEngineConn interface response information is incorrect


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-manager

### What happened + What you expected to happen

The response information of the EngineRestfulApi createEngineConn interface is incorrect, which will cause NPE in the client call. Because NodeStatus is not set to Response
EngineRestfulApi createEngineConn接口response信息有误，会导致客户端调用出现NPE。因为NodeStatus没有设置到Response中导致
![image](https://user-images.githubusercontent.com/11496700/160390685-6d4c8d9d-f026-40c6-961a-52f409c70b9d.png)

### Relevent platform

all

### Reproduction script

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1815, opened 2022-03-28
fix issue that once job npe #1813

casionone:
fix issue that once job npe #1813


peacewong (approved):
LGTM.

Adamyuanyuan:
thanks

Issue #1751, opened 2022-03-18
[Feature] 脚本自定义变量运行类型和后缀约束配置化（Script custom variable run type and suffix constraint configuration）


CCweixiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

新增引擎后，脚本中自定义变量无法正常解析，需要修改如下几个脚本文件：
After the engine is added, the custom variables in the script cannot be parsed normally, and the following script files need to be modified:
`linkis-commons/linkis-storage/src/main/scala/org/apache/linkis/storage/script/parser/QLScriptParser.scala`
`linkis-commons/linkis-storage/src/main/scala/org/apache/linkis/storage/script/compaction/QLScriptCompaction.scala`
`linkis-computation-governance/linkis-entrance/src/main/scala/org/apache/linkis/entrance/interceptor/impl/CustomVariableUtils.scala`

<img width="831" alt="image" src="https://user-images.githubusercontent.com/27730211/158977434-7b67323c-e08b-42c4-83cd-f68630743fff.png">

<img width="585" alt="image" src="https://user-images.githubusercontent.com/27730211/158977464-c8250a22-8fa1-40e8-83ee-1d38de821091.png">

### Description

把上述写死在代码中的配置改为在配置文件中
Change the above configuration written in the code to the configuration file

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1829, opened 2022-03-30
Script custom variable run type and suffix constraint configuration.

CCweixiao:
引擎类型、脚本类型自定义后，脚本中的变量解析异常，需要在源码中修改几个脚本，基于此，需要把此类配置抽取出来作为配置文件中的一个配置项。https://github.com/apache/incubator-linkis/issues/1751

After the engine type and script type are customized, the variable parsing in the script is abnormal, and several scripts need to be modified in the source code. Based on this, such configuration needs to be extracted as a configuration item in the configuration file. https://github.com/apache/incubator-linkis/issues/1751

### Brief change log
增加配置项，以及对应的配置解析工具方法, Add configuration items, and the corresponding configuration parsing tool method

  val CODE_TYPE_AND_RUN_TYPE_RELATION = CommonVars("wds.linkis.storage.codeType.runType.relation", "sql=>hql|sql|jdbc|hive|psql|sh|shell,python=>python|py,java=>java,scala=>scala")

```java
  def getCodeTypeAndRunTypeRelationMap: Map[String, List[String]] = {
    if (CODE_TYPE_AND_RUN_TYPE_RELATION.getValue == null || "".equals(CODE_TYPE_AND_RUN_TYPE_RELATION.getValue)) {
      return Map()
    }

    CODE_TYPE_AND_RUN_TYPE_RELATION.getValue.split(",")
      .filter(x => x != null && !"".equals(x))
      .map(x => {
        val confArr = x.split("=>")
        if (confArr.length == 2) (confArr(0), for (x <- confArr(1).split("\\|").toList) yield x.trim) else null
      }).filter(x => x != null).toMap
  }
```
只是还有几个问题需要确认：
1. 对应配置项CODE_TYPE_AND_RUN_TYPE_RELATION 的定义是否合理，放在StorageConfiguration中是否合适？
2. CustomVariableUtils文件中，runType匹配关系如下
```java
    runType match {
      case "hql" | "sql" | "jdbc" | "hive"| "psql" => codeType = SQL_TYPE
      case "python" | "py" => codeType = PY_TYPE
      case "java" => codeType = JAVA_TYPE
      case "scala" => codeType = SCALA_TYPE
      case "sh" | "shell" => codeType = SQL_TYPE
      case _ => return (false, code)
    }
```
其中sh shell类型对应的codeType SQL_TYPE 这样是否合理？
3. orchestrator模块中有一个文件也叫CustomVariableUtils，其中用到了上述配置，但是缺少linkis-storage模块依赖，
```xml
       <dependency>
            <groupId>org.apache.linkis</groupId>
            <artifactId>linkis-storage</artifactId>
            <version>${linkis.version}</version>
            <scope>provided</scope>
        </dependency>
```
是否可以直接加呢？会不会引起依赖冲突

### DIFF
--- a/linkis-commons/linkis-storage/src/main/scala/org/apache/linkis/storage/script/compaction/PYScriptCompaction.scala
+++ b/linkis-commons/linkis-storage/src/main/scala/org/apache/linkis/storage/script/compaction/PYScriptCompaction.scala
@@ -17,15 +17,12 @@
  
 package org.apache.linkis.storage.script.compaction
 
+import org.apache.linkis.storage.utils.StorageConfiguration
+
 
 class PYScriptCompaction private extends CommonScriptCompaction {
 
-  override def belongTo(suffix: String): Boolean = {
-    suffix match {
-      case "python"|"py"|"sh" => true
-      case _ => false
-    }
-  }
+  override def belongTo(suffix: String): Boolean = StorageConfiguration.getSuffixBelongToRunTypeOrNot(suffix, StorageConfiguration.RUN_TYPE_PYTHON)
 
   override def prefix: String = "#@set"
 
--- a/linkis-commons/linkis-storage/src/main/scala/org/apache/linkis/storage/script/compaction/QLScriptCompaction.scala
+++ b/linkis-commons/linkis-storage/src/main/scala/org/apache/linkis/storage/script/compaction/QLScriptCompaction.scala
@@ -17,16 +17,12 @@
  
 package org.apache.linkis.storage.script.compaction
 
+import org.apache.linkis.storage.utils.StorageConfiguration
 
-class QLScriptCompaction private extends CommonScriptCompaction{
 
-  override def belongTo(suffix: String): Boolean = {
-    suffix match {
-      case "sql" => true
-      case "hql" => true
-      case _ => false
-    }
-  }
+class QLScriptCompaction private extends CommonScriptCompaction {
+
+  override def belongTo(suffix: String): Boolean = StorageConfiguration.getSuffixBelongToRunTypeOrNot(suffix, StorageConfiguration.RUN_TYPE_SQL)
 
   override def prefix: String = "--@set"
 
--- a/linkis-commons/linkis-storage/src/main/scala/org/apache/linkis/storage/script/compaction/ScalaScriptCompaction.scala
+++ b/linkis-commons/linkis-storage/src/main/scala/org/apache/linkis/storage/script/compaction/ScalaScriptCompaction.scala
@@ -17,14 +17,13 @@
  
 package org.apache.linkis.storage.script.compaction
 
+import org.apache.linkis.storage.utils.StorageConfiguration
+
 
 class ScalaScriptCompaction private extends CommonScriptCompaction{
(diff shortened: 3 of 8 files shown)


casionone:
`org.apache.linkis.governance.common.paser.CodeType` has also mapping functions similar to codetype
```
object CodeType extends Enumeration {
  type CodeType = Value
  val Python, SQL, Scala, Shell, Other, Remain, JSON = Value

  def getType(codeType: String): CodeType = codeType.toLowerCase() match {
    case "python" | "pyspark" | "py" => Python
    case "sql" | "hql" | "psql" => SQL
    case "scala" => Scala
    case "shell" | "sh" => Shell 
    case _ => Other
  }
}

```
`"shell" | "sh" => Shell`  is inconsistent with   `case "sh" | "shell" => codeType = SQL_TYPE` ,this also confuses me
I understand that this mapping relationship should be global. Is it more reasonable to put it in the public module of linkis.

casionone:
In addition, I think it is necessary to add unit tests for  
`StorageConfiguration.getCodeTypeAndRunTypeRelationMap`
`StorageConfiguration.getSuffixBelongToRunTypeOrNot`

peacewong:
Thanks.
Question 1: This is reasonable, but I think the CodeTypeAndRunTypeRelationMap should be made a singleton instead of regenerating every time.
问题1：这个是合理的，但是我认为应该将CodeTypeAndRunTypeRelationMap作为单例，而不是每次都重新生成

Question 2 seems to be a bug, it should be classified as a python class or defined as a shell class, because their comments all start with "#", not the "--" of sql
问题2这个看是Bug，应该归类为python类或者定义为shell类，因为他们的注释都是“#”开头，而不是sql的“--”开头

Question 3 CustomVariableUtils in the orchestrator module will be removed, the code is being merged
问题3  orchestrator模块中的CustomVariableUtils将会被移除掉，正在合并代码

CCweixiao:
> 



> In addition, I think it is necessary to add unit tests code for `StorageConfiguration.getCodeTypeAndRunTypeRelationMap` `StorageConfiguration.getSuffixBelongToRunTypeOrNot`

待会的提交，一并带上测试用例

CCweixiao:
> `org.apache.linkis.governance.common.paser.CodeType` has also mapping functions similar to codetype
> 
> ```
> object CodeType extends Enumeration {
>   type CodeType = Value
>   val Python, SQL, Scala, Shell, Other, Remain, JSON = Value
> 
>   def getType(codeType: String): CodeType = codeType.toLowerCase() match {
>     case "python" | "pyspark" | "py" => Python
>     case "sql" | "hql" | "psql" => SQL
>     case "scala" => Scala
>     case "shell" | "sh" => Shell 
>     case _ => Other
>   }
> }
> ```
> 
> `"shell" | "sh" => Shell` is inconsistent with `case "sh" | "shell" => codeType = SQL_TYPE` ,this also confuses me I understand that this mapping relationship should be global. Is it more reasonable to put it in the public module of linkis.

目前只碰到脚本中参数解析的时候，遇到需要在代码中维护这些codeType和runType的变量，我待会看下别的地方是否有相同引用，以做统一

Issue #1816, opened 2022-03-28
[Feature]Maven timeout issues in github builds


casionone:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Sometimes git action of build will compilation fails due to `Connection timed out`. 
failure git action : https://github.com/apache/incubator-linkis/actions?query=is%3Afailure 
```
Error:  Failed to execute goal on project linkis-engineconn-manager-server: Could not resolve dependencies for project org.apache.linkis:linkis-engineconn-manager-server:jar:1.1.0: Could not transfer artifact net.java.dev.jna:jna-platform:jar:5.6.0 from/to central (https://repo.maven.apache.org/maven2): transfer failed for https://repo.maven.apache.org/maven2/net/java/dev/jna/jna-platform/5.6.0/jna-platform-5.6.0.jar: Connection timed out (Read failed) -> [Help 1]
```
From here https://github.com/actions/virtual-environments/issues/1499, we can understand that this is a common problem, I think we can refer to this solution to try to solve  it

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1842, opened 2022-03-31
[build] Fix maven connection timeout issue in gitaction CI

casionone:
[build] Fix maven connection timeout issue in gitaction CI #1816


peacewong (approved):
LGTM.

Issue #1858, opened 2022-04-01
[Feature] [RPC]RPC should support merging of publicservice services


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

RPC should support merging of publicservice services

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1861, opened 2022-04-01
RPC should support merging of publicservice services

peacewong:
#1858

### Brief change log
(for example:)
- RPC should support merging of publicservice services


casionone (approved):
LGTM

PR #1881, opened 2022-04-02
Modify the timeout judgment condition of DevFlinkSQLStreamingListener

maidangdang44:
当使用Flink引擎进行FlinkSQL开发调试时，假设配置的超时时间是10分钟，作业的启动时间为某天的00:00:00 000，按照现在的判断条件，作业会到00:20:00 000才会停止，这样用户在00:10:00 - 00:20:00的时候发现作业没有被终止，会认为系统有问题。
When using the Flink engine for FlinkSQL development and debugging, assume that the configured timeout time is 10 minutes, and the job start time is 00:00:00 000 on a certain day. According to the current judgment conditions, the job will not be stop until 00:20:00 000. So that the user finds that the job is still running at 00:10:00 - 00:20:00, and will think that there is a problem with the system

### Brief change log
修改超时条件。
Modify the timeout condition.


peacewong (approved):
LGTM.

PR #1886, opened 2022-04-02
Supports arbitrary time formatting/Support time operation

duhanmin:
image


casionone:
The code needs to be formatted by running `mvn spotless:apply`
Junt4 is not recommended, use Junt5 uniformly

linkis-computation-governance/linkis-entrance/src/main/java/org/apache/linkis/entrance/utils/FormattedTimeOperationPlaceholders.java
```diff
@@ -0,0 +1,297 @@
...
+import java.util.Date;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ *
+ * 支持任意时间格式化和加减运算
+ * 例如:${yyyyMMdd%-1d}/${yyyy-MM-01%-2M}等等
+ *
+ * Author: duhanmin
```
casionone, line 40: Apache Linkis is licensed under the Apache License Version 2.0, which does not allow to use author tag in  code file header

PR #1898, opened 2022-04-03
When a GatewaySSOUtils user successfully logs in to generate a cookie, it does not directly set the second-level domain name, but adds a configuration switch. The default configuration is false. 

CCweixiao:
新增Cookie domain的设置开关配置，默认值false。
Added the setting switch configuration of Cookie domain, the default value is false


peacewong (approved):
LGTM.

PR #1922, opened 2022-04-07
add junit test for gateway-authentication

casionone:

Add junit test code  for gateway-authentication


peacewong (approved):
LGTM.

Issue #1929, opened 2022-04-08
[Bug] [LinkisManager] Resource check does not check ECM's resources


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-manager

### What happened + What you expected to happen

Resource check does not check ECM's resources

### Relevent platform

all

### Reproduction script

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1930, opened 2022-04-08
Fix Resource check does not check ECM's resources 

peacewong:
#1929

### Brief change log
- Fix Resource check does not check ECM's resources ;


casionone (approved):
LGTM

Issue #1906, opened 2022-04-06
[Bug] [UDF][Python]The python engine incorrectly loads a Pyspark function


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-engineConnplugin

### What happened + What you expected to happen

The python engine incorrectly loads the functions of Pyspark, not the functions of the stand-alone Python itself
python引擎错误加载的是Pyspark的函数，而不是单机Python本身的函数

### Relevent platform

all

### Reproduction script

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1931, opened 2022-04-08
Modify the python engine to not load pyspark functions 

peacewong:
#1906 

### Brief change log
- Modify the python engine to not load pyspark functions


Alexkun (approved):
LGTM.

Issue #1970, opened 2022-04-14
[Feature] [EC]Computation ec asynchronous execution should support setting the number of concurrent job groups


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Computation ec asynchronous execution should support setting the number of concurrent job groups

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1971, opened 2022-04-14
Computation ec asynchronous execution supports setting the number of concurrent job groups

peacewong:
#1970

### Brief change log
- Computation ec asynchronous execution supports setting the number of concurrent job groups


casionone (approved):
LGTM

Issue #1924, opened 2022-04-07
[Bug] [Web] Optimize remove api interface dependency on dss


casionone:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

web

### What happened + What you expected to happen

```
web/src/common/service/apiCommonMethod.js 全部是dss相关的接口 全部移除
web/src/common/config/apiPath.js 全部是dss相关的接口 全部移除
 
web/src/common/service/mixin.js
        api.fetch(`/dss/getAppjointProjectIDByApplicationName`

web/src/apps/URM/module/functionManagement/index.vue
	api.fetch('/dss/framework/workspace/listAllUsers', 'get')
	
	
web/src/apps/URM/module/udfManagement/index.vue
	api.fetch('/dss/framework/workspace/listAllUsers', 'get')
	
web/src/dss/* 这linkis管理台自身是否必须，是否可以考录移除，如果必须保留，名字上是否需要调整 

```
**所有以/dss开始的请求接口都是dss服务的接口，linkis不应该依赖dss服务，需要移除这部分代码。**

### Relevent platform

none

### Reproduction script

none

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

Issue #1888, opened 2022-04-02
[Bug] [Web][1.1.1] dss/getBaseInfo http api with NoApplicationExistsException


casionone:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

web

### What happened + What you expected to happen

![企业微信截图_16488936642933](https://user-images.githubusercontent.com/7869972/161378107-b5b639b4-f3e7-447a-b576-35a4c2eabd72.png)

![企业微信截图_16488933793430](https://user-images.githubusercontent.com/7869972/161377945-62a0594a-9d2d-4600-9806-cf34e75ec8fc.png)

linkis does not depend on dss this interface seems unreasonable. 

And there is an abnormal icon in the top bar

### Relevent platform

none

### Reproduction script

refresh page

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

peacewong:
The management console needs to sort out the interfaces and remove the non-linkis-related interfaces
管理台需要梳理下接口，移除非linkis相关的接口

PR #1972, opened 2022-04-14
Optimize remove api interface dependency on dss

casionone:
Optimize remove api interface dependency on dss

#1924 
#1888


peacewong (approved):
LGTM

Issue #1859, opened 2022-04-01
[Bug] [1.1.1] When calling the saveFullTree interface, configuration not saved on the page will be associated with the wrong label id


husofskyzy:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-public-enhancements

### What happened + What you expected to happen

When calling the saveFullTree interface, the configuration that has not been saved on the page will be associated with the wrong label id. The setting of the default value needs to be optimized.

### Relevent platform

all

### Reproduction script

no

### Anything else

no

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #1984, opened 2022-04-15
Fix Configutration save configuration duplicate Key issue

peacewong:
#1859

### Brief change log
- Fix Configutration save configuration duplicate Key issue
- Optimize code


casionone (approved):
LGTM

Issue #1986, opened 2022-04-15
[Bug] [Entrance][1.1.1]Entrance real-time log gets abnormal number of rows


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-entrance

### What happened + What you expected to happen

The number of lines obtained from the Entrance real-time log is abnormal, and skip skips the number of bytes, not the number of lines. Causes the obtained log to be duplicated
Entrance实时日志获取行数异常，skip跳过的是字节数，不是行数。导致获取的日志重复

### Relevent platform

all

### Reproduction script

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1988, opened 2022-04-15
Fix issue #1986 that Entrance real-time log gets abnormal number of rows

peacewong:
#1986 

### Brief change log
- Fix issue #1986 that Entrance real-time log gets abnormal number of rows;


casionone (approved):
LGTM

Alexkun (approved):
LGTM.

Issue #1990, opened 2022-04-18
[Bug] [Orchestrator][1.1.2]Orchestrator Reheater module has a concurrency issue that makes retries less than configured


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-orchestrator

### What happened + What you expected to happen

Orchestrator Reheater module has a concurrency issue that makes retries less than configured
Orchestrator Reheater模块存在并发问题，使重试次数小于配置的

Now the Reheater steps are:
1. Perform reheater first
2. Retrieve the task for execution
However, there is a concurrency problem. When the reheater is performed in the first step, it is possible that the ExecTask is not completed, and the execution is completed when the second step is performed. This will cause the task to not be reheated. Caused to fail before reaching the number of retries

现在Reheater步骤是：
1. 先进行reheater
2. 再获取任务进行执行
但是存在并发问题，在第一步进行 reheater的时候，有可能ExecTask没有执行完，进行第二步的时候执行完了。这将会导致该任务没有进行Reheater。导致还没达到重试次数就失败了

### Relevent platform

all

### Reproduction script

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1992, opened 2022-04-18
Fix reheater concurrency issue 

peacewong:
Fix reheater concurrency issue #1990

### Brief change log
- Reheater and get task execution are modified to get the same batch of tasks


Alexkun (approved):
LGTM.

PR #2010, opened 2022-04-19
Entrace service：Resolve memory leaks

wpfysztqx:
What is the purpose of the change
https://github.com/apache/incubator-linkis/issues/2009

Brief change log
linkis-cg-entrace service 
Resolve memory leaks


peacewong (approved):
LGTM.

PR #2036, opened 2022-04-26
Fix: ECM page 'instance' display mistake.

Liveipool:
fix:Ecm page 'instance' display mistake

### Brief change log
- Change table's param 'instances' into 'instance' in ECM page.


peacewong (approved):
LGTM.

Issue #2043, opened 2022-04-27
[Feature] [1.1.1] Rollback #1925, the SQL comment clearing logic in SQLCodeParser is duplicated with the SQL comment uninterceptor function


CCweixiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Rollback #1925, the SQL comment clearing logic in SQLCodeParser is duplicated with the SQL comment uninterceptor function

回滚#1925，SQLCodeParser中SQL注释清除逻辑与SQL注释取消拦截器功能重复

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1925, opened 2022-04-07
Fix sql execution fails when comment contains semicolon

CCweixiao:
https://github.com/apache/incubator-linkis/issues/1923

PR #2046, opened 2022-04-27
[Feature-2043] Rollback #1925, the SQL comment clearing logic in SQLCodeParser is duplicated with the SQL comment uninterceptor function

CCweixiao:
https://github.com/apache/incubator-linkis/issues/2043


peacewong (approved):
LGTM.

Issue #2045, opened 2022-04-27
[Feature] [1.1.2] Refactor the matching relationship between script type and running type in the EntranceInterceptor implementation class


CCweixiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

重构EntranceInterceptor实现类中脚本类型和运行类型匹配关系
Refactor the matching relationship between script type and running type in the EntranceInterceptor implementation class

### Description

重构EntranceInterceptor实现类中脚本类型和运行类型匹配关系
Refactor the matching relationship between script type and running type in the EntranceInterceptor implementation class

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2047, opened 2022-04-27
[Feature-2045] Refactor the matching relationship between script type and running type in the EntranceInterceptor implementation class

CCweixiao:
https://github.com/apache/incubator-linkis/issues/2045


peacewong (approved):
LGTM

PR #2067, opened 2022-04-29
Fix ClassCastException of OpenLooKeng Engine

casionone:
```
(Long) EngineConnConf.ENGINE_TASK_EXPIRE_TIME().getValue()
```
fix: ClassCastException when trying to convert Object into java.lang.Long


peacewong (approved):
LGTM.

Issue #2064, opened 2022-04-29
[Feature] For the same field in different tables, the length needs to be kept the same


casionone:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

For the same field in different tables, the length needs to be kept the same 
such as:
```
linkis_ps_udf_user_load udf_id  int(11)->linkis_ps_udf_user_load bigint(20)
linkis_ps_udf_baseinfo  id  bigint(20)
```
There may be other sql statements that have this problem and need to be checked

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #2068, opened 2022-04-29
Optimization of  sql schema for v1.1.1 

casionone:
add upgrade sql schema  for 1.1.1 
for the same field in different tables, the length needs to be kept the same #2064


peacewong (approved):
LGTM.

PR #2070, opened 2022-05-05
fix the ddl bug for openlookeng  

casionone:
Fix the bug of  openlookeng engine  ddl


peacewong (approved):
LGTM.

PR #2087, opened 2022-05-07
UseFormat is required in functionManagement and udfManagement

Liveipool:
UseFormat is required in functionManagement and udfManagement

### Brief change log
- add require logic in functionManagement and udfManagement page


peacewong (approved):
LGTM.

PR #2090, opened 2022-05-07
use python3 the "/appcom/Install/anaconda3/bin/python" not found

Zosimer:
use python3 the "/appcom/Install/anaconda3/bin/python" not found
image

If modified like this, both python and python3 can be used


peacewong (approved):
LGTM.

Issue #2126, opened 2022-05-17
[Bug] 1.2.0 Open file configuration value cannot be parsed properly


CCweixiao:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-public-enhancements

### What happened + What you expected to happen

脚本中只保存配置类型变量时，打开脚本时此配置值无法被解析进metadata中；
脚本中保存自定义变量和运行时配置值时，打开脚本时才能正常解析这两种类型的变量。

When only the configuration type variable is saved in the script, the configuration value cannot be parsed into the metadata when the script is opened;
When custom variables and runtime configuration values ​​are saved in the script, these two types of variables can only be resolved normally when the script is opened.

### Relevent platform

Google brower

### Reproduction script

none

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2127, opened 2022-05-17
[Bug-2126] Fix the bug of open file runtime configuration value cannot be parsed properly.

CCweixiao:
https://github.com/apache/incubator-linkis/issues/2126


peacewong (approved):
LGTM.

PR #2142, opened 2022-05-20
Fix the issue that the JDBC Engine console configuration cannot take effect immediately after modification

huiyuanjjjjuice:
This problem is caused by the JDBC engine caching JDBC engine configuration. The solution is to modify the cache expiration time to configurable

The purpose is to modify the JDBC engine configuration to take effect immediately when the JDBC cache expiration time configuration is set for a short time


peacewong (approved):
LGTM.

Issue #2149, opened 2022-05-22
[Feature] Custom variable logic optimization


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

It is recommended to optimize the logic of custom variables. Now both the common and entry modules have the logic of custom variable substitution. It is recommended to optimize them into the common module.
建议优化自定义变量逻辑，现在在common和entrance模块都有自定义变量替换的逻辑，建议优化到common模块中

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2150, opened 2022-05-22
Custom variable logic optimization 

peacewong:
#2149

### Brief change log
- Custom variable code optimization Unify the variableutils of the entry and common modules to the common module


Alexkun (approved):
LGTM.

Issue #2063, opened 2022-04-29
[Feature] [LinkisManager][1.1.3]LinkisManager record table records ec information


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

LinkisManager record table records ec information, and adds ec instances to linkis_cg_rm_resource_action_record table
1. When the resource is used, it is set by the instance of engineTypeLabel, the method is: ResourceManager.resourceUsed
2. and provide an interface for querying for restful

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2151, opened 2022-05-22
LinkisManager record table records ec information 

peacewong:
#2063

### Brief change log
- LinkisManager records historical ec information
- optimizes support for viewing historical ec logs


Alexkun (approved):
LGTM.

Issue #2167, opened 2022-05-24
[Feature] File type variable configuration in FileSource（FileSource 中文件类型变量配置化）


CCweixiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

```scala
object FileSource {
  private val fileType = Array("dolphin", "sql", "scala", "py", "hql", "python", "out", "log", "text", "sh", "jdbc", "ngql", "psql", "fql")
}
```
对fileType定义的变量加载配置文件配置
Load configuration file configuration for variables defined by fileType

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2168, opened 2022-05-24
[feture-2167] File type variable configuration in FileSource.scala

CCweixiao:
https://github.com/apache/incubator-linkis/issues/2167


peacewong (approved):
LGTM.

Issue #2082, opened 2022-05-06
[Feature] DataSource Manager module optimization


casionone:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

The new features introduced by linkis in 1.1.0. The management of data sources supports users to configure the relevant configuration of data source connections, and by using these connection configurations, create connections and query the metadata data of the corresponding library. Metadata service positioning for data sources is a query service that only provides metadata, and does not involve modifying metadata information. Therefore, the metadata query service module is named: linkis-metedata-manager-server with manager, which is easy to cause misunderstanding and does not match the function positioning. The optimization adjustment is manager->query, that is, linkis-metedata-query-server.
At present, the SDK client LinkisMetaDataRemoteClient for metadata query uses the data source id to associate to obtain the configuration. For upstream users, the data source id attribute is the self-incrementing id of a data table, which is too abstract and has no actual specific meaning. It is recommended to optimize For association through the data source name parameter.

---

目前linkis在1.1.0中引入的新特性。数据源的管理，支持用户配置数据源连接的相关配置，通过使用这些连接配置，创建连接，查询对应库的元信息数据。对于数据源提供元数据服务定位，是只提供元数的查询服务，不涉及到修改元数据信息。所以对于元数据查询服务模块命名为：linkis-metedata-manager-server 带有manager，容易造成误解，与功能定位也不匹配，优化调整为manager->query 即linkis-metedata-query-server.

目前元数据查询的SDK客户端LinkisMetaDataRemoteClient 查询时通过数据源id来关联获取配置，对于上游用户来说，数据源id这个属性是一个数据表的自增id，太抽象，没实际具体含义，建议优化为通过数据源名称参数来作关联。

实现的目标
● 模块命名调整 linkis-metedata-manager-server->linkis-metedata-query-server
● 元数据LinkisMetaDataRemoteClient使用关键入参调整 由数据源id，调整为数据源名称name
实现大体思路
● 1.客户端在组建请求实体类对象时，由原来的dataSourceId，调整为名称dataSourceName
● 2. 相关的后端http接口 入参由原来的dataSourceId调整为dataSourceName，并通过dataSourceName 查询到数据源配置，进行数据库的连接

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2181, opened 2022-05-26
[Feature-2082]DataSource Manager module optimization

casionone:
for #2082

- adjust metadataquery scripts
- optimize of metadata source client
- for compatible with metadata module name refactoring
- add test for MetadataQueryRestful


peacewong (approved):
LGTM.

Issue #2140, opened 2022-05-20
[Feature] [1.2.0] Consistent adjustment of JDBC connection parameters in the JDBC engine.


CCweixiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

JDBC 引擎中JDBC连接参数一致化调整
Consistent adjustment of JDBC connection parameters in the JDBC engine
<img width="1020" alt="image" src="https://user-images.githubusercontent.com/27730211/169474158-5528aa5a-ba86-41cd-a475-74ee676ffb1e.png">

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2185, opened 2022-05-27
[feture-2140] Unify database connection parameters in JDBC engine and…

CCweixiao:
https://github.com/apache/incubator-linkis/issues/2140

Unify database connection parameters in JDBC engine and extract variables, and the unit test connection to local mysql has been tested.


peacewong (approved):
LGTM.

PR #2197, opened 2022-05-30
Adjust linkis-env.sh:modify some configuration defaults

casionone:
Modify some configuration defaults.
```
export ENABLE_HDFS=false 
export ENABLE_HIVE=false
export ENABLE_SPARK=false 
->
export ENABLE_HDFS=true
export ENABLE_HIVE=true
export ENABLE_SPARK=true
```


peacewong (approved):
LGTM.

Issue #2199, opened 2022-05-31
[Feature] Label code optimization remove label keyvalue


peacewong:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Label code optimization removes the persistence of label keyvalue. Now labelVue is not used after persistence. It is recommended to mark it as expired

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2200, opened 2022-05-31
[Feature-2199] Optimize label manager logic to remove label keyValue

peacewong:
#2199 

### Brief change log
- Optimize label manager logic to remove label keyValue


liuyou2 (approved):
LGTM.

Issue #2212, opened 2022-06-02
[Feature] [1.2.0] Add the default DWSClientConfig constructor for LinkisDataSourceRemoteClient to simplify the client API for internal microservices to call data source services


CCweixiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

为LinkisDataSourceRemoteClient增加默认的DWSClientConfig的构造方法，来简化内部微服务调用数据源服务的client API

Add the default DWSClientConfig constructor for LinkisDataSourceRemoteClient to simplify the client API for internal microservices to call data source services

### solutions

Add one method for LinkisDataSourceRemoteClient

```scala
  def createClientConfig(): DWSClientConfig = {
    val serverUrl = Configuration.getGateWayURL()
    if(StringUtils.isEmpty(serverUrl)) throw new DataSourceClientBuilderException("serverUrl cannot be null.")

    val maxConnection: Int = CONNECTION_MAX_SIZE.getValue
    val connectTimeout: Int = CONNECTION_TIMEOUT.getValue
    val readTimeout: Int = CONNECTION_READ_TIMEOUT.getValue
    val tokenKey = AUTH_TOKEN_KEY.getValue
    val tokenValue = AUTH_TOKEN_VALUE.getValue

    val authenticationStrategy: AuthenticationStrategy = new TokenAuthenticationStrategy()
    DWSClientConfigBuilder.newBuilder()
      .addServerUrl(serverUrl)
      .connectionTimeout(connectTimeout)
      .discoveryEnabled(false)
      .discoveryFrequency(1, TimeUnit.MINUTES)
      .loadbalancerEnabled(true)
      .maxConnectionSize(maxConnection)
      .retryEnabled(false)
      .readTimeout(readTimeout)
      .setAuthenticationStrategy(authenticationStrategy)
      .setAuthTokenKey(tokenKey)
      .setAuthTokenValue(tokenValue)
      .setDWSVersion(Configuration.LINKIS_WEB_VERSION.getValue)
      .build()
  }
```

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2213, opened 2022-06-03
[Feature-2212] Add the default DWSClientConfig constructor for LinkisDataSourceRemoteClient to simplify the client API for internal microservices to call data source services

CCweixiao:
https://github.com/apache/incubator-linkis/issues/2212

[Feature-2212] Add the default DWSClientConfig constructor for LinkisDataSourceRemoteClient to simplify the client API for internal microservices to call data source services


peacewong (approved):
LGTM.

PR #2246, opened 2022-06-08
Modify the upgrade sql of version 1.1.3

peacewong:
Modify the upgrade sql of version 1.1.3

### Brief change log
- Modify the upgrade sql of version 1.1.3


Alexkun (approved):
LGTM.

PR #2256, opened 2022-06-09
Fix NPE of client downloadAction

casionone:
fix NPE of client downloadAction


legendtkl:
LGTM

Issue #2255, opened 2022-06-09
[Bug] [1.1.x] The expired field is missing when querying information from a single data source


CCweixiao:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-public-enhancements

### What happened + What you expected to happen

类似/api/rest_j/v1/data-source-manager/info/name/test_mysql2 查询单个数据源信息时，expire属性一直为false，
底层查询sql中未包含expire字段

Similar to /api/rest_j/v1/data-source-manager/info/name/test_mysql2, when querying a single data source information, the expire attribute is always false,
The expire field is not included in the underlying query sql

### Relevent platform

None

### Reproduction script

None

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2261, opened 2022-06-09
[Bug] [1.1.x] The expired field is missing when querying information from a single data source.

CCweixiao:
https://github.com/apache/incubator-linkis/issues/2255
[Bug] [1.1.x] The expired field is missing when querying information from a single data source.


peacewong (approved):
LGTM.

Issue #2191, opened 2022-05-29
[Feature][1.1.3] Add eureka offline apis for services 


legendtkl:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

To support services offline dynamically.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2268, opened 2022-06-12
[Feat][Common] Add common rest api /offline for all services

legendtkl:
issue ref: https://github.com/apache/incubator-linkis/issues/2191

### Brief change log
- add new offline api

### Verify
Before request the api, we can see all the services in Eureka.
image

After we call offline api
image

we can see the service `engineconnmanager` has been offline
image


peacewong (approved):
LGTM.

PR #2278, opened 2022-06-14
Dev 1.2.0 (add test case for jdbc engine)

QuintinTao:
add test case for jdbc engine
- Added tests for JDBCEngineConnExecutor.


CCweixiao (commented):
我看有些测试方法连接的是你自己的mysql，在自己开发环境OK，但在GitHub check时有可能会报错，请再确认下。

I see that some test methods are connected to your own mysql, which is OK in your own development environment, but may report an error when checking on GitHub, please confirm again.

PR #2356, opened 2022-06-28
Dispose of sqoop engine startup failure due to dependency library conflict

hzdhgf:
Dispose of sqoop engine startup failure due to dependency library conflict: log4j-slf4j library conflict causes sqoop engine startup failure, you need to exclude log4j-slf4j-impl library in the imported dependency library linkis-common.


peacewong (approved):
LGTM.

Issue #2273, opened 2022-06-13
[Feature] [1.2.0] Data source edit form supports Chinese and English


CCweixiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Data source edit form supports Chinese and English
数据源编辑表单支持中英文

### Description

<img width="868" alt="wecom-temp-3cb6a1b2529618480eee6b02a0be05de" src="https://user-images.githubusercontent.com/27730211/173362300-5d330ca5-911d-4f26-9c65-b647e41fe9f5.png">

目前，用户切换英文之后，数据源表单添加的label和输入框placeholder文本不支持中英文对应切换
At present, after the user switches to English, the label added to the data source form and the placeholder text of the input box do not support the corresponding switching between Chinese and English

### solutions

```sql
CREATE TABLE `linkis_ps_dm_datasource_type_key`
(
    `id`                  int(11)                       NOT NULL AUTO_INCREMENT,
    `data_source_type_id` int(11)                       NOT NULL,
    `key`                 varchar(32) COLLATE utf8_bin  NOT NULL,
    `name`                varchar(32) COLLATE utf8_bin  NOT NULL,
    `default_value`       varchar(50) COLLATE utf8_bin  NULL     DEFAULT NULL,
    `value_type`          varchar(50) COLLATE utf8_bin  NOT NULL,
    `scope`               varchar(50) COLLATE utf8_bin  NULL     DEFAULT NULL,
    `require`             tinyint(1)                    NULL     DEFAULT 0,
    `description`         varchar(200) COLLATE utf8_bin NULL     DEFAULT NULL,
    `value_regex`         varchar(200) COLLATE utf8_bin NULL     DEFAULT NULL,
    `ref_id`              bigint(20)                    NULL     DEFAULT NULL,
    `ref_value`           varchar(50) COLLATE utf8_bin  NULL     DEFAULT NULL,
    `data_source`         varchar(200) COLLATE utf8_bin NULL     DEFAULT NULL,
    `update_time`         datetime                      NOT NULL DEFAULT CURRENT_TIMESTAMP,
    `create_time`         datetime                      NOT NULL DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin;

```
示例：对应增加字段name_en 和 description_en
[...]

PR #2388, opened 2022-06-30
fix-bug #2273 Data source edit form supports Chinese and English

yyuser5201314:
Data source edit form supports Chinese and English
Related issues: #2273

### DIFF
--- a/linkis-dist/package/db/linkis_ddl.sql
+++ b/linkis-dist/package/db/linkis_ddl.sql
@@ -5,16 +5,16 @@
  * The ASF licenses this file to You under the Apache License, Version 2.0
  * (the "License"); you may not use this file except in compliance with
  * the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


linkis-dist/package/db/linkis_ddl.sql
```diff
@@ -5,16 +5,16 @@
  * The ASF licenses this file to You under the Apache License, Version 2.0
  * (the "License"); you may not use this file except in compliance with
  * the License.  You may obtain a copy of the License at
- * 
+ *
```
CCweixiao, line 8: @peacewong 这种空格对文件造成有修改可以嘛，我看这些应该是不同操作系统的idea自动对文件做了什么格式化操作？

peacewong: ok, it should be fine

linkis-public-enhancements/linkis-datasource/linkis-datasource-manager/server/src/main/java/org/apache/linkis/datasourcemanager/core/dao/mapper/DataSourceParamKeyMapper.xml
```diff
@@ -32,19 +32,42 @@
         <result property="refValue" column="ref_value"/>
         <result property="dataSource" column="data_source" />
     </resultMap>
+    <resultMap id="paramKeyMapEn" type="org.apache.linkis.datasourcemanager.common.domain.DataSourceParamKeyDefinition">
```
CCweixiao, line 35: Is it possible to use the same resultMap and use aliases for query fields?E.g：description_en as description, name_en as name

Issue #2395, opened 2022-07-01
[Bug] Spring-web:SynchronossPartHttpMessageReader should only create temp directory when needed


casionone:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-service-gateway

### What happened + What you expected to happen

The  same as  https://github.com/spring-projects/spring-framework/issues/27092

Some solutions can be shared by referring to this https://blog.csdn.net/ankeway/article/details/86747313
Spring-web has stopped using Synchronoss as default multiparty parser as of version 5.3, where we ship our own DefaultPartHttpMessageReader, which only creates temp directories when the size exceeds a defined limit.

It is necessary to evaluate whether to upgrade the spring-web version separately or upgrade the spring-boot version.
spring-boot 2.3.12.RELEASE is already the last version of 2.3.x

----
官方有这个bug的反馈，且已经修复。

spring-web 从 5.3 版开始，已经停止使用 Synchronoss 作为默认的多方解析器，在其中发布了自己的DefaultPartHttpMessageReader，它仅在大小超过定义的限制时创建临时目录

具体的处理方案可以参考  
- https://blog.csdn.net/ankeway/article/details/86747313
- https://www.cnblogs.com/cndarren/p/15982972.html

需要评估下是单独升级spring-web版本（spring-web 5.2.15.RELEASE 到5.2.16.RELEASE），还是升级spring-boot版本。spring-boot 2.3.12.RELEASE 已经是2.3.x的最后一个版本

### Relevent platform

none

### Reproduction script

none

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

duhanmin:
https://github.com/apache/incubator-linkis/pull/2397
编译下看看

PR #2397, opened 2022-07-01
[ISSUE-2395]SynchronossPartHttpMessageReader should only create temp directory when needed/CVE-2022-22965

duhanmin:
https://github.com/spring-projects/spring-framework/issues/27092
1 . SynchronossPartHttpMessageReader should only create temp directory when needed 
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-22965
2. A Spring MVC or Spring WebFlux application running on JDK 9+ may be vulnerable to remote code execution (RCE) via data binding. The specific exploit requires the application to run on Tomcat as a WAR deployment. If the application is deployed as a Spring Boot executable jar, i.e. the default, it is not vulnerable to the exploit. However, the nature of the vulnerability is more general, and there may be other ways to exploit it.

- Dependencies (does it add or upgrade a dependency): (yes)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,7 +102,7 @@
         <linkis.version>1.1.3</linkis.version>
         <hadoop.version>2.7.2</hadoop.version>
         <zookeeper.version>3.5.9</zookeeper.version>
-        <spring.boot.version>2.3.12.RELEASE</spring.boot.version>
+        <spring.boot.version>2.6.9</spring.boot.version>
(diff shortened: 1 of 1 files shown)


duhanmin:
https://github.com/apache/incubator-linkis/issues/2395

jackxu2011:
this issue should only upgrade spring framework not spring boot

Issue #2389, opened 2022-06-30
[Bug] linkis-ps-metadataquery module is missing after compiling the code


yyuser5201314:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-public-enhancements

### What happened + What you expected to happen

After [Improvement]Optimize of linkis assamble-combined-package module #2374 merging the code on branch dev-1.2.0,
The linkis-ps-metadataquery module is missing after compiling the code.
This problem may be due to confusion between the module linkis-metadata-manager and the module linkis-metadata-query
This problem also causes the installation and startup scripts to fail

### Relevent platform

dev-1.2.0

### Reproduction script

![捕获](https://user-images.githubusercontent.com/63859058/176748124-80ecf323-af39-4259-89fd-41415a170508.PNG)
![捕获2](https://user-images.githubusercontent.com/63859058/176748548-6390ad3a-d2ae-464f-b9ae-ce85d1412dce.PNG)

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

casionone:
I did make a mistake when doing this optimization,
You may need to submit a fix for this issue as soon as possible.
If you are not sure about the PR submission process, you can refer to this guide doc https://github.com/apache/incubator-linkis/issues/1161

PR #2398, opened 2022-07-02
[ISSUE 2389] Fixed the bug in the linkis-ps-metadataquery module

yyuser5201314:
Fixed the bug in the linkis-ps-metadataquery module that was missing after packaging
Related issues: #2389


yyuser5201314:
Other branches may have the same problem

casionone:
image
linkis-dist/package/sbin/graceful-upgrade.sh  need to be modified about  `metadatamanager`

Issue #2414, opened 2022-07-05
[Bug] SimpleDateFormat is thread-unsafe, causing problems such as task state not flipping properly


huiyuanjjjjuice:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-public-enhancements

### What happened + What you expected to happen

1、SimpleDateFormat is thread-unsafe, causing problems such as task state not flipping properly
2、For example,when a large number of tasks are submitted, the UpdateTimeMills property of the JobHistory class is incorrect because the SimpleDateFormat thread is not safe. As a result, the linkis_ps_job_history_group_history table cannot update its status properly

### Relevent platform

none

### Reproduction script

Problems can be reproduced during task stress testing

### Anything else

none

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2418, opened 2022-07-05
Fixed the problem that the task state could not be flipped normally when thread unsafe SimpleDateFormat was used as a global variable

huiyuanjjjjuice:
Fixed the problem that the task state could not be flipped normally when thread unsafe SimpleDateFormat was used as a global variable

For example:
1、when a large number of tasks are submitted, the UpdateTimeMills property of the JobHistory class is incorrect because the SimpleDateFormat thread is not safe. As a result, the linkis_ps_job_history_group_history table cannot update its status properly

2、Problems can be reproduced during task stress testing

### DIFF
--- a/linkis-engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/imexport/CsvRelation.scala
+++ b/linkis-engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/imexport/CsvRelation.scala
@@ -56,10 +56,11 @@ class CsvRelation(@transient private val source: Map[String, Any]) extends Seria
   val quote = LoadData.getMapValue[String](source, "quote", "\"")
   val escape = LoadData.getMapValue[String](source, "escape", "\\")
   val escapeQuotes = LoadData.getMapValue[Boolean](source, "escapeQuotes", false)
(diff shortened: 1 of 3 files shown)


huiyuanjjjjuice:
#2414

linkis-engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/imexport/CsvRelation.scala
```diff
@@ -56,8 +56,14 @@ class CsvRelation(@transient private val source: Map[String, Any]) extends Seria
   val quote = LoadData.getMapValue[String](source, "quote", "\"")
   val escape = LoadData.getMapValue[String](source, "escape", "\\")
   val escapeQuotes = LoadData.getMapValue[Boolean](source, "escapeQuotes", false)
-  val dateFormat: SimpleDateFormat = new SimpleDateFormat(LoadData.getMapValue[String](source, "dateFormat", "yyyy-MM-dd"), Locale.US)
-  val timestampFormat: SimpleDateFormat = new SimpleDateFormat(LoadData.getMapValue[String](source, "timestampFormat", "yyyy-mm-dd hh:mm:ss"), Locale.US)
+
+  private val dateFormatLocal = new ThreadLocal[SimpleDateFormat]() {
```
peacewong, line 60: dateFormatLocal can be removed

peacewong (approved):
LGTM.

Issue #1763, opened 2022-03-21
[Bug] API /linkisManager/rm/applicationlist has NPE bug


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-manager

### What happened + What you expected to happen

![image](https://user-images.githubusercontent.com/11496700/159197939-1f075f9e-ec4a-48f1-910d-8bf03ee54515.png)

### Relevent platform

Class: RMMonitorRest
```
@RequestMapping(path = Array("applicationlist"), method = Array(RequestMethod.POST))
  def getApplicationList(request: HttpServletRequest, @RequestBody param: util.Map[String, AnyRef]): Message

 val applicationList = creatorToApplicationList(userCreatorLabel.getCreator)
          applicationList.put("usedResource", applicationList("usedResource").asInstanceOf[Resource] + node.getNodeResource.getUsedResource)
          applicationList.put("maxResource", applicationList("maxResource").asInstanceOf[Resource] + node.getNodeResource.getMaxResource)
          applicationList.put("minResource", applicationList("minResource").asInstanceOf[Resource] + node.getNodeResource.getMinResource)
          applicationList.put("lockedResource", applicationList("lockedResource").asInstanceOf[Resource] + node.getNodeResource.getLockedResource)
```
The above code does not make a null value judgment during the asInstance

### Reproduction script

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2440, opened 2022-07-08
[ISSUE-1763] Add non empty validation to the applicationlist interface

binbinCheng:
Add non empty validation to the applicationlist interface
Related issue #1763


casionone (approved):
LGTM

Issue #2450, opened 2022-07-10
[Feature] [1.2.0] The new storage path is optimized when the engine material is updated


CCweixiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

wds.linkis.bml.hdfs.prefix
wds.linkis.bml.local.prefix
如果用户有修改了这两个配置，那么引擎物料文件需要更新时，获取到的旧物料文件路径可能不可用导致引擎物料无法完成更新。
If the user has modified these two configurations, when the engine material file needs to be updated, the obtained old material file path may not be available and the engine material cannot be updated.

### Description

原有VersionServiceImpl类中的updateVersion方法内，生成新路径的逻辑是：
In the updateVersion method in the original VersionServiceImpl class, the logic for generating a new path is:
```java
// create resource new path
String newVersion = params.get("newVersion").toString();
String path = versionDao.getResourcePath(resourceId) + "_" + newVersion;
```
对于生成后的path，如果bml prefix有改动，需要重新成功一个path。
For the generated path, if the bml prefix is changed, a new path needs to be succeeded.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2451, opened 2022-07-10
[Feature-2450] The new storage path is optimized when the engine material is updated.

CCweixiao:
https://github.com/apache/incubator-linkis/issues/2450


peacewong (approved):
LGTM.

PR #2466, opened 2022-07-14
[ISSUE-2475]linkis-bml-server fix package name capitalization problem

binbinCheng:
org.apache.linkis.bml.Entity Package name modification
Related issue #2475


casionone:
linkis-dist/package/conf/linkis-ps-publicservice.properties  
`wds.linkis.server.mybatis.typeAliasesPackage` need to be changed together

PR #2477, opened 2022-07-15
Optimize build out dependence of linkis-cg-engineconnplugin 

casionone:
Optimize build out dependence of linkis-cg-engineconnplugin   linkis-cg-engineconnmanager


peacewong (approved):
LGTM.

PR #2483, opened 2022-07-18
Add install front-end dependencies

utopianet:
Add install front-end dependencies


legendtkl:
LGTM, thanks for the job.

legendtkl:
It seems Licence check failed because of file `web/.env`, would you help fix this by the way. @utopianet

PR #2489, opened 2022-07-18
Fix the problem that 'CsvRelation' class cannot be serialized

### LINKED ISSUE #2488: [Bug]The CsvRelation class cannot be serialized because of the ThreadLocal object, which will cause an error in importing CSV files into the hive table
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-cg-engineConnplugin

### What happened + What you expected to happen

1、The CsvRelation class cannot be serialized because of the ThreadLocal object, which will cause an error in importing CSV files into the hive table

2、The fallback code will solve this problem

### Relevent platform

 linkis

### Reproduction script

 none

### Anything else

none

huiyuanjjjjuice:
1、CsvRelation cannot be serialized due to the ThreadLocal object
2、Roll back the code to resolve the problem
Fixes #2488


peacewong (approved):
LGTM.

Issue #2529, opened 2022-07-26
[Feature]linkis-public-enhancements/linkis-bml/linkis-bml-server unit test


binbinCheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

linkis-public-enhancements/linkis-bml/linkis-bml-server unit test

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #2530, opened 2022-07-26
[ISSUES-2529] bml unit test

binbinCheng:
linkis-public-enhancements/linkis-bml/linkis-bml-server  unit test
Related issue #2529


casionone:
The asf header needs to be added at the beginning of the file

Issue #2549, opened 2022-08-01
Task execution can be very slow when printing a long result set to the console


huiyuanjjjjuice:
For example, if the python engine print has a long result set, the task takes a long time to execute

PR #2550, opened 2022-08-01
Optimize the read efficiency of the result set when the result set is output after the script is executed

huiyuanjjjjuice:
#2549


peacewong (approved):
LGTM.

Issue #2559, opened 2022-08-03
[Bug]  Variable substitution error


duhanmin:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-commons

### What happened + What you expected to happen

Variable substitution error

```
dss_autotest.demo_data{ds=20220516}
```

Incorrect results
```
dss_autotest.demo_data228=20220516
```

### Relevent platform

-

### Reproduction script

-

### Anything else

-

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2560, opened 2022-08-03
Variable substitution error(#2559)

duhanmin:
* del {***} Variable substitution, Only support ${***}
* ${} replace with #{}
* #{} reduction of priority

https://github.com/apache/incubator-linkis/issues/2559


peacewong (approved):
LGTM.

PR #2600, opened 2022-08-04
ES engine  code optimization and NoSuchMethodError bug fix

aiceflower:
1. code optimization：Remove white space at both ends of the string #2604 
2. NoSuchMethodError bug fix: add OPENLOOKENG,TRINO,ELASTICSEARCH case #2603 

### Brief change log
- modify org.apache.linkis.manager.label.entity.engine.EngineType.
- modify org.apache.linkis.engineplugin.elasticsearch.executer.client.EsClientFactory.

### DIFF
--- a/linkis-computation-governance/linkis-manager/linkis-label-common/src/main/java/org/apache/linkis/manager/label/entity/engine/EngineType.scala
+++ b/linkis-computation-governance/linkis-manager/linkis-label-common/src/main/java/org/apache/linkis/manager/label/entity/engine/EngineType.scala
@@ -27,7 +27,7 @@ object EngineType extends Enumeration with Logging {
 
   val SPARK = Value("spark")
 
-  val HIVE  = Value("hive")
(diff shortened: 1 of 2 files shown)


peacewong (approved):
LGTM.

linkis-computation-governance/linkis-manager/linkis-label-common/src/main/java/org/apache/linkis/manager/label/entity/engine/EngineType.scala
```diff
@@ -92,6 +92,9 @@ object EngineType extends Enumeration with Logging {
     case _ if APPCONN.toString.equals(str) => APPCONN
```
casionone, line 92: I don't think the scalability is so good in this way. 
try to use Enumeration.withName replace it 

https://stackoverflow.com/questions/33593525/scala-safe-way-of-converting-string-to-enumeration-value

PR #2629, opened 2022-08-08
EC's Yarn app id log adjusted to the file yarnApp.log file

### LINKED ISSUE #2628: [Feature] EC Yarn app id logs should be printed to a separate log
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

The EC Yarn app id log should be printed in a separate log, because stderr may have a lot of logs, resulting in performance problems with the appid filter

### Description

_No response_

### Use case

_No response_

### solutions

_No response_

### Anything else

_No response_

peacewong:
close #2628 

### Brief change log
- EC's Yarn app id log adjusted to the file yarnApp.log file


casionone:
Is it not necessary to format the code for the api class files involved in swagger? After formatting, the code is pulled longer, and the readability will be poor. 

de3108ee1808c7ec28e0d0d8c2313d5

Issue #2632, opened 2022-08-08
[Bug] The rollbackversion function modifies the case


binbinCheng:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-public-enhancements

### Steps to reproduce

![image](https://user-images.githubusercontent.com/106590848/183398047-34560dbf-5128-47d0-9490-8fad92052d36.png)

### Expected behavior

URL capitalization causes request 404

### Your environment

- Linkis version used: dev-1.2.0

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #2633, opened 2022-08-08
[ISSUES-2632]The rollbackversion function modifies the case

binbinCheng:
The rollbackversion function modifies the case
Related issue #2632


peacewong (approved):
LGTM.

PR #2654, opened 2022-08-09
fix test case for ShellEngineConnExecutor

QuintinTao:
modify test case for ShellEngineConnExecutor
add  System.setProperty("HADOOP_CONF_DIR", "./")


peacewong (approved):
LGTM.

Issue #2739, opened 2022-08-12
[Bug] The result set may not be completely written to HDFS


huiyuanjjjjuice:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-commons

### Steps to reproduce

For example, when printing a large data set using python, the result set in HDFS is partially missing because it is not fully written

### Expected behavior

Result set write complete

### Your environment

 Linkis version used: 1.0.3

### Anything else

none

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2740, opened 2022-08-12
Fix the problem that the result set is not completely written to HDFS

huiyuanjjjjuice:
For example, when printing a large data set using python, the result set in HDFS is partially missing because it is not fully written

#2739


peacewong (approved):
LGTM.

Issue #2876, opened 2022-08-18
[linkis-gateway-httpclient-support]Scala code format alarm clear 


binbinCheng:
Scala code format alarm clear[linkis-gateway-httpclient-support]
---
scala code format alarm clearing task
Adjust the code format and clear the warn alarm in the idea

The scala code formatting rules configuration file is located in scalastyle-config.xml
You can check for format exceptions by running the command mvn scalastyle:check under the module

Module linkis-spring-cloud-services/linkis-service-gateway/linkis-gateway-httpclient-support

--- 

scala 代码格式告警清除任务
调整代码格式，清除idea中的warn告警

scala 代码格式规则配置文件位于 scalastyle-config.xml
你可以通过模块下 运行命令mvn scalastyle:check 检查格式异常

模块 linkis-spring-cloud-services/linkis-service-gateway/linkis-gateway-httpclient-support

![image](https://user-images.githubusercontent.com/106590848/185333936-02559d2e-9545-4d57-9505-d100260eb794.png)

 | class | path| 
| ------------- |------------- |
| DWSHttpClient|org.apache.linkis.httpclient.dws.DWSHttpClient| 
| DWSResult|org.apache.linkis.httpclient.dws.response.DWSResult|
| DWSHeartbeatResult| org.apache.linkis.httpclient.dws.response.DWSHeartbeatResult|
| DWSAuthenticationResult| org.apache.linkis.httpclient.dws.response.DWSAuthenticationResult|
| UnknownVersionException| org.apache.linkis.httpclient.dws.exception.UnknownVersionException|
| StaticAuthenticationStrategy|org.apache.linkis.httpclient.dws.authentication.StaticAuthenticationStrategy|

PR #3130, opened 2022-08-30
[linkis-gateway-httpclient-support] Modification of scala file floati…

binbinCheng:
[linkis-gateway-httpclient-support] Modification of scala file floating red
Related issues: #2876


peacewong (approved):
LGTM.

PR #3132, opened 2022-08-30
[linkis-variable] Modification of scala file floating red

binbinCheng:
[linkis-variable] Modification of scala file floating red
Related issues: https://github.com/apache/incubator-linkis/issues/2873


peacewong (approved):
LGTM.

Issue #2872, opened 2022-08-18
[linkis-udf-client]Scala code format alarm clear


binbinCheng:
Scala code format alarm clear[linkis-udf-client]
---
scala code format alarm clearing task
Adjust the code format and clear the warn alarm in the idea

The scala code formatting rules configuration file is located in scalastyle-config.xml
You can check for format exceptions by running the command mvn scalastyle:check under the module

Module linkis-public-enhancements/linkis-publicservice/linkis-udf/linkis-udf-client

--- 

scala 代码格式告警清除任务
调整代码格式，清除idea中的warn告警

scala 代码格式规则配置文件位于 scalastyle-config.xml
你可以通过模块下 运行命令mvn scalastyle:check 检查格式异常

模块 linkis-public-enhancements/linkis-publicservice/linkis-udf/linkis-udf-client

![image](https://user-images.githubusercontent.com/106590848/185317727-4e2880b6-ac38-4270-8601-d0ef82fcbb7e.png)

 | class | path| 
| ------------- |------------- |
| UDFClient|org.apache.linkis.udf.UDFClient|

PR #3133, opened 2022-08-30
[linkis-udf-client] Modification of scala file floating red

binbinCheng:
[linkis-udf-client] Modification of scala file floating red
Related issues: https://github.com/apache/incubator-linkis/issues/2872


peacewong (approved):
LGTM.

Issue #2875, opened 2022-08-18
[linkis-gateway-core]Scala code format alarm clear 


binbinCheng:
Scala code format alarm clear[linkis-gateway-core]
---
scala code format alarm clearing task
Adjust the code format and clear the warn alarm in the idea

The scala code formatting rules configuration file is located in scalastyle-config.xml
You can check for format exceptions by running the command mvn scalastyle:check under the module

Module linkis-spring-cloud-services/linkis-service-gateway/linkis-gateway-core

--- 

scala 代码格式告警清除任务
调整代码格式，清除idea中的warn告警

scala 代码格式规则配置文件位于 scalastyle-config.xml
你可以通过模块下 运行命令mvn scalastyle:check 检查格式异常

模块 linkis-spring-cloud-services/linkis-service-gateway/linkis-gateway-core

![image](https://user-images.githubusercontent.com/106590848/185324210-ae73507a-61f0-4e1d-aa1d-7016b532e949.png)

 | class | path| 
| ------------- |------------- |
| UserRestful|org.apache.linkis.gateway.security.UserRestful| 
| SecurityFilter|org.apache.linkis.gateway.security.SecurityFilter|
| ProxyUserUtils| org.apache.linkis.gateway.security.ProxyUserUtils|
| LinkisPreFilter|org.apache.linkis.gateway.security.LinkisPreFilter|
| LDAPUserRestful| org.apache.linkis.gateway.security.LDAPUserRestful|
| GatewaySSOUtils| org.apache.linkis.gateway.security.GatewaySSOUtils|
| TokenAuthentication|org.apache.linkis.gateway.security.token.TokenAuthentication|
| SSOInterceptor | org.apache.linkis.gateway.security.sso.SSOInterceptor | 
| GatewayRouter | org.apache.linkis.gateway.route.GatewayRouter | 
| GatewayParser | org.apache.linkis.gateway.parser.GatewayParser |

PR #3153, opened 2022-08-31
[linkis-gateway-core] Modification of scala file floating red

binbinCheng:
[linkis-gateway-core] Modification of scala file floating red
Related issues: https://github.com/apache/incubator-linkis/issues/2875


peacewong (approved):
LGTM.

Issue #2859, opened 2022-08-18
[linkis-bml-server]Scala code format alarm clear


binbinCheng:
Scala code format alarm clear[linkis-bml-server]
---
scala code format alarm clearing task
Adjust the code format and clear the warn alarm in the idea

The scala code formatting rules configuration file is located in scalastyle-config.xml
You can check for format exceptions by running the command mvn scalastyle:check under the module

Module linkis-public-enhancements/linkis-bml/linkis-bml-server

--- 

scala 代码格式告警清除任务
调整代码格式，清除idea中的warn告警

scala 代码格式规则配置文件位于 scalastyle-config.xml
你可以通过模块下 运行命令mvn scalastyle:check 检查格式异常

模块 linkis-public-enhancements/linkis-bml/linkis-bml-server

![image](https://user-images.githubusercontent.com/106590848/185279296-7eb76d29-9efc-4642-9f3f-d98354020449.png)

 | class | path| 
| ------------- |------------- |
| BmlReceiver|org.apache.linkis.bml.rpc.BmlReceiver| 
| BmlResourceExpiredException|org.apache.linkis.bml.common.BmlResourceExpiredException|
| BmlAuthorityException| org.apache.linkis.bml.common.BmlAuthorityException|

PR #3165, opened 2022-08-31
[linkis-bml-server] Modification of scala file floating red

binbinCheng:
[linkis-bml-server] Modification of scala file floating red
Related issues: https://github.com/apache/incubator-linkis/issues/2859


peacewong (approved):
LGTM.

Issue #2856, opened 2022-08-18
[linkis-orchestrator-core]Scala code format alarm clear


binbinCheng:
Scala code format alarm clear[linkis-orchestrator-core]
---
scala code format alarm clearing task
Adjust the code format and clear the warn alarm in the idea

The scala code formatting rules configuration file is located in scalastyle-config.xml
You can check for format exceptions by running the command mvn scalastyle:check under the module

Module linkis-orchestrator/linkis-orchestrator-core

--- 

scala 代码格式告警清除任务
调整代码格式，清除idea中的warn告警

scala 代码格式规则配置文件位于 scalastyle-config.xml
你可以通过模块下 运行命令mvn scalastyle:check 检查格式异常

模块 linkis-orchestrator/linkis-orchestrator-core

![image](https://user-images.githubusercontent.com/106590848/185092464-d13b2214-4f2b-4e37-89ec-5236bff9838b.png)

 | class | path| 
| ------------- |------------- |
| Orchestrator|org.apache.linkis.orchestrator.Orchestrator| 
| StatusInfoExecTask|org.apache.linkis.orchestrator.strategy.StatusInfoExecTask|
| ResultSetExecTask|org.apache.linkis.orchestrator.strategy.ResultSetExecTask|
| AsyncExecTaskRunnerImpl| org.apache.linkis.orchestrator.strategy.async.AsyncExecTaskRunnerImpl|
| AbstractReheater| org.apache.linkis.orchestrator.reheater.AbstractReheater|
| CodeLogicalUnit| org.apache.linkis.orchestrator.plans.unit.CodeLogicalUnit|
| JobExecTask|org.apache.linkis.orchestrator.plans.physical.JobExecTask|
| Origin| org.apache.linkis.orchestrator.plans.logical.Origin | 
| JobTask | org.apache.linkis.orchestrator.plans.logical.JobTask | 
| AbstractTask | org.apache.linkis.orchestrator.plans.logical.AbstractTask |
| RuntimeParams|org.apache.linkis.orchestrator.plans.ast.RuntimeParams| 
| QueryParams|org.apache.linkis.orchestrator.plans.ast.QueryParams|
| AbstractStage|org.apache.linkis.orchestrator.plans.ast.AbstractStage|
| AbstractJob| org.apache.linkis.orchestrator.plans.ast.AbstractJob|
| OrchestratorListenerBusContext| org.apache.linkis.orchestrator.listener.OrchestratorListenerBusContext|
[...]

PR #3184, opened 2022-09-01
[linkis-orchestrator-core] Modification of scala file floating red

binbinCheng:
[linkis-orchestrator-core] Modification of scala file floating red
Related issues: https://github.com/apache/incubator-linkis/issues/2856


casionone (approved):
LGTM

Issue #2849, opened 2022-08-17
[linkis-computation-orchestrator]Scala code format alarm clear


binbinCheng:
Scala code format alarm clear[fllinkis-computation-orchestratornk]
---
scala code format alarm clearing task
Adjust the code format and clear the warn alarm in the idea

The scala code formatting rules configuration file is located in scalastyle-config.xml
You can check for format exceptions by running the command mvn scalastyle:check under the module

Module linkis-orchestrator/linkis-computation-orchestrator

--- 

scala 代码格式告警清除任务
调整代码格式，清除idea中的warn告警

scala 代码格式规则配置文件位于 scalastyle-config.xml
你可以通过模块下 运行命令mvn scalastyle:check 检查格式异常

模块 linkis-orchestrator/linkis-computation-orchestrator

![image](https://user-images.githubusercontent.com/106590848/185086994-2986f72f-7cae-419f-bbbb-e832a95f311a.png)

 | class | path| 
| ------------- |------------- |
| ComputationOrchestratorSessionFactory | org.apache.linkis.orchestrator.computation.ComputationOrchestratorSessionFactory| 
| TreeNodeUtil|org.apache.linkis.orchestrator.computation.utils.TreeNodeUtil|
| ComputationTaskExecutionReceiver| org.apache.linkis.orchestrator.computation.service.ComputationTaskExecutionReceiver|
| CacheExecTask| org.apache.linkis.orchestrator.computation.physical.CacheExecTask|
| CodeLogicalUnitExecTask| org.apache.linkis.orchestrator.computation.physical.CodeLogicalUnitExecTask|
| AbstractProgressOperation| org.apache.linkis.orchestrator.computation.operation.progress.AbstractProgressOperation|
| EngineConnMonitor|org.apache.linkis.orchestrator.computation.monitor.EngineConnMonitor|
| CodeExecTaskExecutorManager| org.apache.linkis.orchestrator.computation.execute.CodeExecTaskExecutorManager| 
| DefaultCodeExecTaskExecutorManager |org.apache.linkis.orchestrator.computation.execute.DefaultCodeExecTaskExecutorManager | 
| ComputationJobReq | org.apache.linkis.orchestrator.computation.entity.ComputationJobReq |
| LabelRegularCheckRuler| org.apache.linkis.orchestrator.computation.catalyst.validator.LabelRegularCheckRuler| 
[...]

PR #3187, opened 2022-09-01
[linkis-computation-orchestrator] Modification of scala file floating…

binbinCheng:
[linkis-computation-orchestrator] Modification of scala file floating red
Related issues: https://github.com/apache/incubator-linkis/issues/2849


casionone (approved):
LGTM

Issue #2836, opened 2022-08-17
[io_file]Scala code format alarm clear


binbinCheng:
Scala code format alarm clear[io_file]
---
scala code format alarm clearing task
Adjust the code format and clear the warn alarm in the idea

The scala code formatting rules configuration file is located in scalastyle-config.xml
You can check for format exceptions by running the command mvn scalastyle:check under the module

Module linkis-engineconn-plugins/io_file

--- 

scala 代码格式告警清除任务
调整代码格式，清除idea中的warn告警

scala 代码格式规则配置文件位于 scalastyle-config.xml
你可以通过模块下 运行命令mvn scalastyle:check 检查格式异常

模块 linkis-engineconn-plugins/io_file

![image](https://user-images.githubusercontent.com/106590848/185061928-021950ce-30e3-4959-b821-51b481091c87.png)

 | class | path| 
| ------------- |------------- |
| IoEngineConnPlugin|org.apache.linkis.manager.engineplugin.io.IoEngineConnPlugin| 
| IOHelp|org.apache.linkis.manager.engineplugin.io.utils.IOHelp|
| FlinkEngineConnResourceFactory | org.apache.linkis.manager.engineplugin.io.service.FsProxyService|
| IoEngineConnExecutor| org.apache.linkis.manager.engineplugin.io.executor.IoEngineConnExecutor|
| FSInfo| org.apache.linkis.manager.engineplugin.io.domain.FSInfo|

PR #3216, opened 2022-09-02
[linkis-engineconn-plugin-flink] Modification of scala file floating red

binbinCheng:
[linkis-engineconn-plugin-flink] Modification of scala file floating red
Related issues: https://github.com/apache/incubator-linkis/issues/2836


casionone (approved):
LGTM

Issue #3226, opened 2022-09-03
[Bug] Set default values for non-null fields in table 'linkis_ps_dm_datasource_type_key'


yyuser5201314:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-dist

### Steps to reproduce

'linkis_ps_dm_datasource_type_key' 表的‘neme_en’字段非空，插入语句必须设置默认值
 
The 'linkis_ps_dm_datasource_type_key' table non-null field must have a default value

### Expected behavior

The 'linkis_ps_dm_datasource_type_key' table non-null field must have a default value

### Your environment

dev-1.3.1

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #3229, opened 2022-09-03
The default values are set for 'description_en' fields in the `linkis_ps_dm_datasource_type_key` table

yyuser5201314:
The default values are set for 'description_en' fields in the `linkis_ps_dm_datasource_type_key` table
Related issues: #3226


casionone (approved):
LGTM

PR #3239, opened 2022-09-04
[Feature] [JDBC-Engine] Optimize JDBC parameter acquisition logic.

CCweixiao:
Optimize JDBC parameter acquisition logic.


peacewong (approved):
LGTM.

PR #3261, opened 2022-09-06
Test case for spark executor

QuintinTao:
test cases for spark executor

### DIFF
--- a/linkis-engineconn-plugins/spark/pom.xml
+++ b/linkis-engineconn-plugins/spark/pom.xml
@@ -377,6 +377,12 @@
       <version>3.0.9</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.poi</groupId>
+      <artifactId>poi</artifactId>
+      <version>5.2.2</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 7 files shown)


linkis-engineconn-plugins/spark/src/test/scala/org/apache/linkis/engineplugin/spark/config/TestSparkConfiguration.scala
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.linkis.engineplugin.spark.config
+
+import org.junit.jupiter.api.{Assertions, Test}
+
+class TestSparkConfiguration {
+
```
peacewong, line 23: It is recommended to remove the configuration test of the configuration, because it is common to modify the configuration parameters

linkis-engineconn-plugins/spark/pom.xml
```diff
@@ -377,6 +377,12 @@
       <version>3.0.9</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.poi</groupId>
```
peacewong, line 381: Introduced in the linkis-storage module, is it necessary here?

Issue #3287, opened 2022-09-08
Fix Insert function in InstanceLabelMapper.xml


binbinCheng:
**Fix Insert function in InstanceLabelMapper.xml**

![企业微信截图_16626070146047](https://user-images.githubusercontent.com/106590848/189027815-9331ecbe-7760-46cc-be8d-5a908a0a0e7e.png)

PR #3288, opened 2022-09-08
Fix Insert function in InstanceLabelMapper.xml

binbinCheng:
Fix Insert function in InstanceLabelMapper.xml
Related issues: #3287


casionone (approved):
LGTM

Issue #3290, opened 2022-09-08
The resultMap return type is not added to some functions in InstanceLabelMapper


binbinCheng:
**The resultMap return type is not added to some functions in InstanceLabelMapper**

![image](https://user-images.githubusercontent.com/106590848/189053302-e2fd8d60-9f75-4e1b-9e4e-7d878cdc1f37.png)

PR #3293, opened 2022-09-08
The resultMap return type is not added to some functions in InstanceLabelMapper 

binbinCheng:
The resultMap return type is not added to some functions in InstanceLabelMapper 
Related issues: https://github.com/apache/incubator-linkis/issues/3290


casionone (approved):
LGTM

Issue #3306, opened 2022-09-09
[bug] auth path NPE problem


aiceflower:
NPE problems may occur when invoked externally
![image](https://user-images.githubusercontent.com/22620332/189267672-081d77b0-01f0-4dd2-9c05-88eb32a8614d.png)

PR #3307, opened 2022-09-09
[ISSUE-3306] Fix  FsPath null String 

aiceflower:
fix NPE problem https://github.com/apache/incubator-linkis/issues/3306


casionone (approved):
LGTM

PR #3428, opened 2022-09-15
Test case for spark executor

QuintinTao:
Test case for spark executor
I followed [peacewong](https://github.com/peacewong)'s opinion.
#3261


peacewong (approved):
LGTM.

PR #3462, opened 2022-09-19
fix publish-snapshot.yml

binbinCheng:
fix publish-snapshot.yml

Related issues：https://github.com/apache/incubator-linkis/issues/3463


casionone (approved):
LGTM

Issue #3494, opened 2022-09-21
Move Java classes in scala into Java packages


binbinCheng:
**Move Java classes in scala into Java packages**

![企业微信截图_16637514592703](https://user-images.githubusercontent.com/106590848/191467446-a5887180-6e2e-4834-9b33-7eb73b401565.png)

PR #3495, opened 2022-09-21
Move Java classes in scala into Java packages

binbinCheng:
**Move Java classes in scala into Java packages**
Related issues: https://github.com/apache/incubator-linkis/issues/3494


casionone (approved):
LGTM

Issue #3508, opened 2022-09-22
Change the docker script file format


aiceflower:
Change the docker script file to the UNIX format

PR #3513, opened 2022-09-22
change docker scripts file format

aiceflower:
modify file format https://github.com/apache/incubator-linkis/issues/3508


jackxu2011 (approved):
LGTM

Issue #3410, opened 2022-09-15
[ linkis-cs-server ]module errorcode optimization and documentation


binbinCheng:
**[linkis-cs-server ]module errorcode optimization and documentation**

**Example：**

**documentation（文档）**
![image](https://user-images.githubusercontent.com/106590848/190046858-6e3eb7ac-fe7e-4cc1-8470-d3e26f59dcb2.png)

**Error code optimization（错误码优化）**

![image](https://user-images.githubusercontent.com/106590848/190352298-e6c5c8f5-5bac-4632-a667-5015475902a2.png)

**Classes that need to be optimized（需要优化的类）**
 | class | path| 
| ------------- |------------- |
| ContextSearchFailedException|org.apache.linkis.cs.exception.ContextSearchFailedException|

PR #3520, opened 2022-09-23
[ISSUE-3410][linkis-cs-server]errorcode code optimization

binbinCheng:
**[linkis-cs-server]errorcode code optimization**
Related issues: #3410


peacewong (approved):
LGTM.

PR #3647, opened 2022-10-13
[ResourceDao]The sql annotation form is optimized to XML 

binbinCheng:
**The sql annotation form is optimized to XML**

Related issues: #3646


linkis-public-enhancements/linkis-bml/linkis-bml-server/src/main/java/org/apache/linkis/bml/dao/ResourceDao.java
```diff
@@ -36,17 +34,12 @@ public interface ResourceDao {
 
   long uploadResource(Resource resource);
 
-  @Select(
-      "select exists(select * from `linkis_ps_bml_resources` where resource_id = #{resourceId}  and enable_flag = 1 )")
-  int checkExists(@Param("resourceId") String resourceId);
+  Integer checkExists(@Param("resourceId") String resourceId);
```
casionone, line 38: exists function can return  null ?

jacktao007 (approved):
LGTM.

PR #3849, opened 2022-11-17
linkis-cli-application - change output when there is no result

binbinCheng:

**Fix the error of cli null result**


peacewong (approved):
LGTM.

PR #3851, opened 2022-11-17
linkis-storage - fix npe bug in VariableParser

binbinCheng:

**The entrance config configuration value is null NPE repair**


peacewong (approved):
LGTM.

Issue #3853, opened 2022-11-18
[Feature] Enabling kerberos authentication by default 


aiceflower:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Enabling kerberos authentication by default 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #3854, opened 2022-11-18
[feat-3853] Enabling kerberos authentication by default

aiceflower:
Enabling kerberos authentication by default https://github.com/apache/incubator-linkis/issues/3853


casionone:
just need to update docs about  https://linkis.apache.org/zh-CN/docs/latest/deployment/deploy-quick

PR #3914, opened 2022-11-28
Update linkis-start-all.sh

utopianet:
The service startup waiting time is set to 15 seconds because many community users have received feedback that the service startup is abnormal.


casionone:
@aiceflower help review

aiceflower:
Waiting 15 seconds after each service starts feels too slow. 
You are advised to wait 15 seconds after the ps-publicservice starts.

jacktao007 (approved):
LGTM.

casionone:
> do we also need to give user a kind notice ?,keep user informed of our `strategy sleep`. for 15s seems a bit long

yes,  maybe echo some notice info before sleep？

GuoPhilipse:
> 

seems fine for me, i have fired a ticker for this improvement. #3946

Issue #3974, opened 2022-12-06
[Bug] The bml resource usage permission is incorrect


aiceflower:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-public-enhancements

### Steps to reproduce

After user A uploals the streamis project file resource, user B references the project file resource in the import task. The reference succeeds. There is a permission problem.

### Expected behavior

User B cannot access the resources uploaded by user A.

### Your environment

linkis dev-1.3.1

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #3977, opened 2022-12-07
[ISSUE-3974] The bml resource usage permission is incorrect

aiceflower:
fix the problem of bml resource usage permission https://github.com/apache/incubator-linkis/issues/3974


peacewong (approved):
LGTM.

Issue #3980, opened 2022-12-08
[Feature]  Linkis Client added the default retry mechanism


aiceflower:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

Linkis Client added the default retry mechanism

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #3981, opened 2022-12-08
[feat-3980] Add default retry strategy

aiceflower:
Linkis Client add default retry strategy https://github.com/apache/incubator-linkis/issues/3980


peacewong (approved):
LGTM.

PR #4100, opened 2023-01-05
[feat:4099] gateway log ip info

aiceflower:
gateway print ip log info https://github.com/apache/linkis/issues/4099


linkis-spring-cloud-services/linkis-service-gateway/linkis-spring-cloud-gateway/src/main/java/org/apache/linkis/gateway/springcloud/http/GatewayAuthorizationFilter.java
```diff
@@ -159,10 +159,14 @@ private Mono<Void> gatewayDeal(
     ServiceInstance serviceInstance;
     try {
       parser.parse(gatewayContext);
+      logger.info("client request uri: " + gatewayContext.getRequest().getRequestURI());
```
guoshupei, line 162: If you want to know the requestURI, I think it would be better to add it at the beginning of GatewayAuthorizationFilter.
Whether the level can be changed to debug？

linkis-spring-cloud-services/linkis-service-gateway/linkis-spring-cloud-gateway/src/main/java/org/apache/linkis/gateway/springcloud/http/GatewayAuthorizationFilter.java
```diff
@@ -159,10 +159,14 @@ private Mono<Void> gatewayDeal(
...
       if (gatewayHttpResponse.isCommitted()) {
         return gatewayHttpResponse.getResponseMono();
       }
       serviceInstance = router.route(gatewayContext);
+      if (serviceInstance != null) {
+        logger.info("route uri: " + serviceInstance.getInstance());
+      }
```
guoshupei, line 169: The route method is already printed and contains more information, I think it can be removed

aiceflower: I put the internal log on the outer layer so that it is easy to locate the problem in multiple threads.

peacewong (approved):
LGTM.

PR #4145, opened 2023-02-05
Echo error message to console when initing db in install.sh

QuantumXiecao:
When I run install.sh in MySQL 5.5 environment, I encountered the following situation: install.sh indicates success but when I run start-all.sh it turns out to be that some tables in linkis_ddl.sh fail to be installed actually. My PR is to echo error message to console when initing db in install.sh. The reason for the issue is that returning code zero of mysql -e "**"  does not mean the source **.sql actually succeeded as depicted by the picture below:
image

The result of My PR would be like as follows:
image

### Brief change log

- Echo error message to console when initing db in install.sh.

### Checklist

- [x] **If this is a code change**: I have fully verified the new behavior.

### DIFF
--- a/linkis-dist/bin/common.sh
+++ b/linkis-dist/bin/common.sh
@@ -69,14 +69,22 @@ function copyFile(){
 
 }
 
-function isSuccess(){
-if [ $? -ne 0 ]; then
+function echoErrMsgAndExit() {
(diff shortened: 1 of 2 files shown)


peacewong (approved):
LGTM.

Issue #2393, opened 2022-07-01
[Bug] [Flink] Not support YarnPerJobClusterDescriptorAdapter for FlinkSQLComputationExecutor


wbmialiu:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar issues.

### Linkis Component

linkis-commons

### What happened + What you expected to happen

当 linkis 有启用个人函数功能，streamis 运行oncejob sql任务时，会导致Not support YarnPerJobClusterDescriptorAdapter for FlinkSQLComputationExecutor 的异常。
![image](https://user-images.githubusercontent.com/7869972/176843901-f5e7ba72-2f05-46b1-941a-2aecaf6df77a.png)

legend91325:
I want to fix this issues，you can assign it to me

legend91325:
> 

why just remove this `wds.linkis.engine.connector.hooks` property ,can fix this bug ?

QuantumXiecao:
I want to fix this issues，you can assign it to me. @jackxu2011 @casionone

PR #4155, opened 2023-02-07
[ISSUE-2393]Remove JarUdfEngineHook from flink engine conn properties

QuantumXiecao:

[ISSUE-2393]Remove JarUdfEngineHook from flink engine conn properties.

After revisiting codes constantly, I have finally got the point: flink engine conn now do not support udf.
So configing the JarUdfEngineHook would cause the issue.
By the way, I think in the future we should also support udf for flink in linkis.

### Related issues/PRs

Related issues: #2393 

### Brief change log

- [ISSUE-2393]Remove JarUdfEngineHook from flink engine conn properties.

### Checklist

- [x] **If this is a code change**: I have fully verified the new behavior.

### DIFF
--- a/linkis-engineconn-plugins/flink/src/main/resources/linkis-engineconn.properties
+++ b/linkis-engineconn-plugins/flink/src/main/resources/linkis-engineconn.properties
@@ -18,5 +18,5 @@ wds.linkis.server.version=v1
 wds.linkis.engineconn.debug.enable=true
 #wds.linkis.keytab.enable=true
 wds.linkis.engineconn.plugin.default.class=org.apache.linkis.engineconnplugin.flink.FlinkEngineConnPlugin
-wds.linkis.engine.connector.hooks=org.apache.linkis.engineconn.computation.executor.hook.ComputationEngineConnHook,org.apache.linkis.engineconn.computation.executor.hook.JarUdfEngineHook
+wds.linkis.engine.connector.hooks=org.apache.linkis.engineconn.computation.executor.hook.ComputationEngineConnHook
 wds.linkis.engineconn.executor.manager.class=org.apache.linkis.engineconnplugin.flink.executormanager.FlinkExecutorManager


linkis-computation-governance/linkis-engineconn/linkis-computation-engineconn/src/main/scala/org/apache/linkis/engineconn/computation/executor/hook/UDFLoadEngineConnHook.scala
```diff
@@ -174,11 +174,11 @@ abstract class UDFLoad extends Logging {
     logger.info(s"Successful to execute function code ${runType}, type : ${udfType}")
   }
 
-  protected def loadUDF(labels: Array[Label[_]]): Unit = {
+  protected def loadUDF(): Unit = {
 
     val codes = getFunctionCode()
     if (null != codes && codes.nonEmpty) {
-      val executor = ExecutorManager.getInstance.getExecutorByLabels(labels)
+      val executor = ExecutorManager.getInstance.getReportExecutor
```
peacewong, line 181: The Executor here needs to be obtained through Labels, and the default Executor cannot be used, because UDF supports different languages. For Spark, there are different UDF registration languages such as SQL, Python, and Scala.

QuantumXiecao: Many thx for your remind! 
I infer from the code that AbstractEngineConnExecution has already created one executor by labels as depicted by codes below:
```
override def getReportExecutor: OnceExecutor = if (getExecutors.isEmpty) {
    val labels =
      if (null == engineConn.getEngineCreationContext.getLabels()) {
        throw new OnceEngineConnErrorException(12560, "No labels is exists.")
      } else engineConn.getEngineCreationContext.getLabels()
    createExecutor(
      engineConn.getEngineCreationContext,
      labels.toArray[Label[_]](Array.empty[Label[_]])
    ).asInstanceOf[OnceExecutor]
  } else {
    getExecutors
      .filter(_.isInstanceOf[OnceExecutor])
      .maxBy {
        case executor: SensibleExecutor => executor.getStatus.ordinal()
        case executor: Executor => executor.getId.hashCode
      }
      .asInstanceOf[OnceExecutor]
  }
```
After creation, the following execution hooks(such as UDFLoadEngineConnHook) should do register UDF stuff. Just as the issue illustrated, the hook created another executor which throws exception to hinder EngineConnServer to start successfully. My PR wants the hook to reuse the executor created before(The previously created executor has already been created by labels). For Spark, different UDF registration languages such as SQL, Python, and Scala must be with different labels. So I should correct my codes again=> UDFLoadEngineConnHook should get executor with full labels instead of just by codeLanguageLabel(which would cause the issue). Am I right? @peacewong

QuantumXiecao: @peacewong I have recorrected my PR according to what I have explained above.

peacewong: @QuantumXiecao Thanks for your update, now the Spark EngineConnUDF registration will register all types of UDFs when EC starts. Because Spark supports multiple Executors, for example, Python Executors can use UDFs registered with Sql and Scala

QuantumXiecao: @peacewong 
After revisiting codes constantly, I have finally got the point: flink engine conn now do not support udf.
So configing the JarUdfEngineHook would cause the issue.
By the way, I think in the future we should also support udf for flink in linkis.

peacewong: @QuantumXiecao Yes, this has been planned into the RoadMap in 2023, now you can remove the corresponding hook

peacewong (approved):
LGTM.

PR #4240, opened 2023-02-17
Modify the default resource limit default value

utopianet:
Modify the default resource limit parameter value to the maximum to avoid that the job cannot run because the resource limit is too small.


peacewong (approved):
LGTM.

PR #4350, opened 2023-03-09
fix(engineconn): fix the java option memory unit bug

jackxu2011:

When the exchangis submit job, with JAVA_ENGINE_REQUEST_MEMORY `1024MB`,  the ec startup failed  

### Checklist

- [ ] I have explained the changes or the new features added to this PR
- [ ] I have verified that this change is backward compatible (If not, please discuss on the [Linkis mailing list](https://linkis.apache.org/community/how-to-subscribe) first)
- [ ] **If this is a code change**: I have written unit tests to fully verify the new behavior.

### DIFF
--- a/linkis-computation-governance/linkis-engineconn/linkis-engineconn-plugin-core/src/main/scala/org/apache/linkis/manager/engineplugin/common/util/NodeResourceUtils.scala
+++ b/linkis-computation-governance/linkis-engineconn/linkis-engineconn-plugin-core/src/main/scala/org/apache/linkis/manager/engineplugin/common/util/NodeResourceUtils.scala
@@ -23,9 +23,14 @@ import org.apache.linkis.manager.engineplugin.common.conf.EngineConnPluginConf
 import org.apache.commons.lang3.StringUtils
 
 import java.util
-import java.util.Objects
+import java.util.{Locale, Objects}
 
 object NodeResourceUtils {
+
+  val JAVA_MEMORY_UNIT_G = "G"
+  val JAVA_MEMORY_UNIT_M = "M"
+  val JAVA_MEMORY_UNIT_K = "K"
+  val JAVA_MEMORY_UNIT_B = "B"
 
   def appendMemoryUnitIfMissing(properties: util.Map[String, String]): Unit = {
     Objects.requireNonNull(properties);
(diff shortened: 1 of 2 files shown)


linkis-computation-governance/linkis-engineconn/linkis-engineconn-plugin-core/src/main/scala/org/apache/linkis/manager/engineplugin/common/util/NodeResourceUtils.scala
```diff
@@ -35,18 +40,35 @@ object NodeResourceUtils {
...
+  def formatJavaOptionMemory(memory: String, defaultUnit: String): String = {
+    assert(StringUtils.isNotBlank(memory))
+    val memoryTmp = memory.toUpperCase(Locale.getDefault)
+    if (
+        memoryTmp.endsWith(JAVA_MEMORY_UNIT_G) || memoryTmp.endsWith(
+          JAVA_MEMORY_UNIT_M
+        ) || memoryTmp.endsWith(JAVA_MEMORY_UNIT_K)
+    ) {
+      memoryTmp
+    } else if (memoryTmp.endsWith(JAVA_MEMORY_UNIT_B)) {
```
casionone, line 61: Xmx not support  unit MB,
If the value unit set by the user  not within the range（G/M/K/B）, whether to consider print  warn log and set it as the default memory 1g

jackxu2011: This is a compatiable with exchangis, if use GB, MB, KB, B,  just drop it.  is this ok?

casionone: This is okay. If the user accidentally enters the wrong unit such as GK, do we need to consider using the default value or add some warn logs

jackxu2011: in this case,  the check should use regex to check?   if the unit wit `GK`, the ec will startup failed.

casionone: > in this case, the check should use regex to check? if the unit wit `GK`, the ec will startup failed.

I think it's ok

casionone (approved):

PR #4357, opened 2023-03-10
Optimization of upload file interface in FsRestfulApi.java

binbinCheng:

**Add relevant default attributes to upload file interface optimization in FsRestfulApi.java**

This configuration is used in the checkIsUsersDirectory function, which is false by default. If the user does not set it, the function will not be able to perform path verification（checkIsUsersDirectory 函数中用到该配置，该配置默认为false，如果用户不设置，该函数将无法进行路径校验）

image

image

image


casionone:
Can you add a more detailed reason for doing this?

peacewong (approved):
LGTM.

Issue #3427, opened 2022-09-15
[Feature][udf]  Load specific udfs as we needed


rarexixi:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-linkis/issues) and found no similar feature requirement.

### Problem Description

 We need to load specific udfs in the following cases:
1. The executeOnce label of the engine is true
2. Test a new udf

### Description

Load specific udfs by specifying the udf‘s id list

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #4417, opened 2023-03-27
[bug][udf] load user's and shared udf

rarexixi:

Load user's and shared udf with ids

### Related issues/PRs

Related issues: #3427

### Brief change log

Load user's and shared udf with ids


peacewong (approved):
LGTM.

PR #4486, opened 2023-04-25
Support tidb dataSources

ChengJie1053:
Support tidb dataSources


peacewong (commented):
need to add update sql in dir package/db/upgrade

ChengJie1053:
> need to add update sql in dir package/db/upgrade我需要在目录包/db/upgrade中添加更新sql

Ok, thanks

PR #4526, opened 2023-05-08
Translate engineconn-plugins-trino service classes from Scala to Java 

ChengJie1053:
Translate engineconn-plugins-trino service classes from Scala to Java


linkis-engineconn-plugins/trino/src/main/java/org/apache/linkis/engineplugin/trino/executor/TrinoEngineConnExecutor.java
```diff
@@ -0,0 +1,558 @@
...
+            String[] rowArray = row.stream().map(r -> String.valueOf(r)).toArray(String[]::new);
+            resultSetWriter.addRecord(new TableRecord(rowArray));
+            rows += 1;
+          }
+        }
+        engineExecutorContext.pushProgress(progress(taskId), getProgressInfo(taskId));
+        statement.advance();
+      }
+    } catch (Exception e) {
+      IOUtils.closeQuietly(resultSetWriter);
```
peacewong, line 453: need to throw e

ChengJie1053: > need to throw e

Ok, thank you for reviewing the code

Issue #4557, opened 2023-05-18
[Bug] If the input script is json, the loop cannot be completed


duhanmin:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/linkis/issues) and found no similar issues.

### Linkis Component

linkis-commons

### Steps to reproduce

https://github.com/apache/linkis/edit/dev-1.4.0/linkis-commons/linkis-common/src/main/java/org/apache/linkis/common/utils/VariableOperationUtils.java

![image](https://github.com/apache/linkis/assets/28647031/9350c71a-e5cf-4753-8e97-c8eceed2f645)

 If the input script is json, the loop cannot be completed

如果输入脚本是json,循环将无法结束

### Expected behavior

rt

### Your environment

- Linkis version used: 1.1.2
- Environment name and version:
    - cdh-5.14.2
    - hdp-3.1.5
    - hive-2.1.1
    - spark-3.2.1
    - scala-2.12.2
    - jdk 1.8.0_121
    - ....

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #4558, opened 2023-05-18
[Fix Bug]If the input script is json, the loop cannot be completed(#4557)

duhanmin:
If the input script is json, the loop cannot be completed(https://github.com/apache/linkis/issues/4557)


peacewong (approved):
LGTM.

peacewong (approved):
LGTM

Issue #4582, opened 2023-05-29
[Feature][linkis-computation-client] Provide a method which can get CreateEngineResult when submit a [once] job


rarexixi:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/linkis/issues) and found no similar feature requirement.

### Solutions

Provide a method which can get CreateEngineResult when submit a [once] job

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #4583, opened 2023-05-29
[Feature][linkis-computation-client] Provide a method which can get CreateEngineResult when submit a [once] job

rarexixi:

Provide a method which can get CreateEngineResult when submit a [once] job

### Related issues/PRs

Related issues: #4582


peacewong (approved):
LGTM.

PR #4588, opened 2023-05-30
Generate spark sql from kafka es mongo

ChengJie1053:
Generate spark sql from kafka es mongo


linkis-public-enhancements/linkis-datasource/linkis-metadata-query/server/src/main/java/org/apache/linkis/metadata/query/server/service/impl/MetadataQueryServiceImpl.java
```diff
@@ -379,32 +381,126 @@ public GenerateSqlInfo getSparkSqlByDsNameAndEnvId(
...
+        String sqlConnectUrl =
+            invokeMetaMethod(
+                dsInfoResponse.getDsType(),
+                "getSqlConnectUrl",
+                new Object[] {dsInfoResponse.getCreator(), dsInfoResponse.getParams()},
+                String.class);
+
+        return getSparkSqlByJdbc(
+            database, table, dsInfoResponse.getParams(), columns, sqlConnectUrl);
+      } else if (dsInfoResponse.getDsType().equalsIgnoreCase("kafka")) {
```
peacewong, line 413: Is it better to define these as constants or enumerations

ChengJie1053: Ok, thank you for reviewing the code

peacewong (approved):
LGTM.

PR #4609, opened 2023-06-07
remove v prefix for impala and add default compile mode

aiceflower:
1.remove v prefix for impala
2.add impala is not compiled by default
To compile impala ec, perform the following operations
 mvn package -P compile-impala


peacewong (approved):
LGTM.

PR #4619, opened 2023-06-12
update check and install scripts

aiceflower:
1. update default hive check version to 3.x
2. add the switch for generating random tokens


peacewong (approved):
LGTM.

PR #4638, opened 2023-06-15
[Fix Bug]Fix spark mongo etl task execution failed

ChengJie1053:
image

The mongo version of the Mongo-Spark-connector is the same as the Mongo-Java-driver version


peacewong (approved):
LGTM.

Issue #4664, opened 2023-06-19
[Bug] [EC]Use database hook should default set false


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/linkis/issues) and found no similar issues.

### Linkis Component

linkis-computation-governance, linkis-engineconnn-plugin

### Steps to reproduce

1. run hql 
2. hook print error log
```
2023-06-19 14:15:56.191 [ERROR] [main                                    ] o.a.h.h.q.Driver (1250) [printError] [JobId-] - FAILED: SemanticException [Error 10072]: Databa
se does not exist: hadoop_ind
org.apache.hadoop.hive.ql.parse.SemanticException: Database does not exist: hadoop_ind
        at org.apache.hadoop.hive.ql.parse.BaseSemanticAnalyzer.getDatabase(BaseSemanticAnalyzer.java:2171)
        at org.apache.hadoop.hive.ql.parse.DDLSemanticAnalyzer.analyzeSwitchDatabase(DDLSemanticAnalyzer.java:1413)
        at org.apache.hadoop.hive.ql.parse.DDLSemanticAnalyzer.analyzeInternal(DDLSemanticAnalyzer.java:516)
        at org.apache.hadoop.hive.ql.parse.BaseSemanticAnalyzer.analyze(BaseSemanticAnalyzer.java:285)
        at org.apache.hadoop.hive.ql.Driver.compile(Driver.java:659)
        at org.apache.hadoop.hive.ql.Driver.compileInternal(Driver.java:1826)
        at org.apache.hadoop.hive.ql.Driver.compileAndRespond(Driver.java:1773)
        at org.apache.hadoop.hive.ql.Driver.compileAndRespond(Driver.java:1768)
        at org.apache.hadoop.hive.ql.reexec.ReExecDriver.compileAndRespond(ReExecDriver.java:126)
        at org.apache.hadoop.hive.ql.reexec.ReExecDriver.run(ReExecDriver.java:214)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #4665, opened 2023-06-19
Use database hook suhould default set false 

peacewong:

Use database hook suhould default set false 

### Related issues/PRs

Related issues: #4664


casionone:
LGTM

casionone (approved):

PR #4746, opened 2023-07-07
The dss prod label interface forwarding problem was fixed.

ahaoyao:

Match prod's interface to prod's service


peacewong (approved):
LGTM.

PR #4895, opened 2023-09-04
fix spark k8s bug

ChengJie1053:
fix spark k8s bug 

1.Caused by: java.nio.file.NoSuchFileException: ~/.kube/config
2.sparkApplication.getStatus() is null


peacewong (approved):
LGTM.

Issue #4911, opened 2023-09-25
[Bug] The sortByResource method in ResourceNodeSelectRule violates the general contract of comparators


sjgllgh:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/linkis/issues) and found no similar issues.

### Linkis Component

linkis-computation-governance

### Steps to reproduce

1. When there are many deployed ECMs and the concurrency is high, an exception may be thrown: Comparison method violates its general contract!
2. The comparison method provided by the code is: when memory, CPU, and the number of instances are all large, return -1; when memory, CPU, and the number of instances are all equal, return 0; in all other cases, return 1.
3. There may be situations where, when the memory of A is larger than that of B, and the CPU of B is larger than that of A, the comparison between A and B will always return 1.

### Expected behavior

Modify the rules of the comparator to prevent violating the general contract of the comparator.

### Your environment

- Linkis version used: 1.4.0
- Environment name and version:
    - jdk 1.8.0_144

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #4915, opened 2023-09-25
[BUG] Fix em sort error

sjgllgh:

Fix the bug that the sortByResource method in ResourceNodeSelectRule violates the general contract of comparators.

### Related issues/PRs

Related issues: #4911

### Brief change log

- Added and implemented a compare method for the Resource abstract class and its subclasses.;
- Modified the Java class ResourceNodeSelectRule to ensure that sortByResource uses the correct sorting method.

### Note
- For LoadInstanceResource, memory is compared first. If the memory is the same, the number of CPU cores is compared. When the number of CPU cores is the same, the number of instances that have been launched will be compared. Other subcategories also adopted similar comparison rules.

### Checklist

- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] **If this is a code change**: I have written unit tests to fully verify the new behavior.


peacewong (approved):
LGTM.

Issue #4916, opened 2023-09-26
[Bug] When submitting tasks concurrently, there may be a problem of exceeding the em threshold


sjgllgh:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/linkis/issues) and found no similar issues.

### Linkis Component

linkis-computation-governance

### Steps to reproduce

1. Submit tasks in batches；
2. Locked resources may exceed the threshold of em.

### Expected behavior

Used resources + locked resources do not exceed the threshold of em

### Your environment

- Linkis version used: 1.4.0

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #4917, opened 2023-09-26
[BUG] when submitting tasks in batches, em resources may exceed the threshold

sjgllgh:

Fix bug: when submitting tasks in batches, em resources may exceed the threshold.

### Related issues/PRs

Related issues: #4916 

### Brief change log

- Modify the DefaultResourceManager.java to fix the bug.

### Checklist

- [x] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] **If this is a code change**: I have written unit tests to fully verify the new behavior.

### DIFF
--- a/linkis-computation-governance/linkis-manager/linkis-application-manager/src/main/java/org/apache/linkis/manager/rm/service/impl/DefaultResourceManager.java
+++ b/linkis-computation-governance/linkis-manager/linkis-application-manager/src/main/java/org/apache/linkis/manager/rm/service/impl/DefaultResourceManager.java
@@ -250,19 +250,6 @@ public class DefaultResourceManager extends ResourceManager implements Initializ
     CombinedLabel combinedLabel = labelContainer.getCombinedResourceLabel();
 
     try {
-      // check ecm resource if not enough return
-      try {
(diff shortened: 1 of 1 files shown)


sjgllgh:
Thank you for your answer. Different scenarios may have different considerations. 

> The logical judgment here is considered as follows:
> 
> 1. The ECM tag is not locked first for resource judgment. If the resource is judged, it will continue. Therefore, there is a certain probability of negative resources, but there will not be too many situations.
> 2. The reason for this consideration is to reduce lock waiting, because here is a concurrent lock grab, so the lock granularity should be as small as possible to reduce concurrency pressure.
> 3. If the usercreator lock is placed at a higher level, it will cause more lock waits.
> 
> 这里的逻辑判断是这样考虑的：
> 
> 1. ECM标签先不锁进行资源判断如果资源判断过就继续，所以存在一定概率出现负资源，但是不会出现太多情况
> 2. 之所以这样考虑是为了减少锁的等待，因为这里是并发抢锁的，所以锁的粒度尽量小减少并发压力
> 3. 如果把usercreator的锁放到更上一层，会导致更多的锁等待

Issue #4851, opened 2023-08-04
[Bug] [Storage] The Linkis result set should support the return of null strings


peacewong:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/linkis/issues) and found no similar issues.

### Linkis Component

linkis-commons

### Steps to reproduce

1. run sql ：`select "null" as c1,"NULL" as c2, cast (NULL as string) as c3;`
2. get result 
`["null","NULL","NULL"]`
3. should to get res :
`["null","NULL",null]`

Now the result of Linkis returns the uppercase string NULL for the null string, which will make it difficult for the downstream system to distinguish the null value of the string.
The plan is amended as follows:

When the result set is stored, the null is stored as linkis_null, and when the result set is read, it is judged that if it is read by DSS, the returned value is still NULL
If it is accessed by other systems, it returns null, and supports replacement by nullValue
Test concerns:
Pay attention to the DSS result set showing null as NULL
Pay attention to the result set download csv, Excel is NULL
Pay attention to whether the task of changing the upstream result to null in Visualis is correct
Focus on exporting the result set to csv, and displaying Excel results as NULL

现在Linkis的结果返回对于字符串的null返回的是大写的字符串NULL，这会导致下游系统不好区分字符串的null值。
本次计划修改为如下：
1. 结果集存储时，对于null存储为linkis_null，结果集读取的时候做判断如果是DSS读取则返回的仍然是NULL
2. 如果是其他系统访问则返回的是null，并且支持通过nullValue进行替换
测试关注点：
1. 关注DSS结果集展示null为NULL
2. 关注结果集下载csv、Excel为NULL
3. 关注Visualis的可是换上游结果null的任务是否正确
4. 关注结果集导出为csv、Excel的结果展示为NULL

### Expected behavior

3. should to get res :
`["null","NULL",null]`

### Your environment

- Linkis version used: 1.1.2
- Environment name and version:
    - cdh-5.14.2
    - hdp-3.1.5
    - hive-2.1.1
    - spark-3.2.1
    - scala-2.12.2
    - jdk 1.8.0_121
    - ....

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #4920, opened 2023-10-03
[Bug] [Storage] The Linkis result set should support the return of null

peacewong:

1. client support return null
2. storage support store null
3. httpClient add default retry for 401

### Related issues/PRs

Related issues: #4851


casionone (approved):
LGTM

Issue #5016, opened 2023-12-05
[Bug] fix gc log


yangwenzea:
### Search before asking

- [X] I searched the [issues](https://github.com/apache/linkis/issues) and found no similar issues.

### Linkis Component

linkis-engineconnn-plugin

### Steps to reproduce

启动fink任务之后跳转yarn 界面 查看env.java.opts参数 
env.java.opts: "-Xloggc:\<LOG_DIR\>/gc.log" 

### Expected behavior

env.java.opts: "-Xloggc:<LOG_DIR>/gc.log "

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #5025, opened 2023-12-06
flink load default configuration

### LINKED ISSUE #5012: [Feature][1.5.0] flink loads env.java.opts configuration
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/linkis/issues) and found no similar feature requirement.

### Problem Description

The current streaming Flink job does not record GC logs, making it impossible to view GC related information. In order to solve this problem, the flink engine loads the env.java.opts configuration in flink-config.yaml. The parameter with the same name submitted by the streamis job has a higher priority than the env.java.opts value of the default configuration file, and must be deduplicated. 

### Description

flink loads env.java.opts configuration

### Use case

_No response_

### Solutions

In FlinkisEngineConnFactory, first read the default flink-config.yaml file in the service and obtain the value of the env.java.opts parameter.
Get the user's env.java.opts, split it first by spaces, and then by equal signs. If there are duplicates, replace them with the user's values. If there are no duplicates, add them after the parameters.

### Anything else

_No response_

yangwenzea:
closes #5012


yangwenzea:
closes #5016

PR #5036, opened 2023-12-12
Fix repl get task result bug

ChengJie1053:
Fix repl get task result bug

### DIFF
--- a/linkis-engineconn-plugins/repl/src/main/java/org/apache/linkis/engineplugin/repl/executor/ReplEngineConnExecutor.java
+++ b/linkis-engineconn-plugins/repl/src/main/java/org/apache/linkis/engineplugin/repl/executor/ReplEngineConnExecutor.java
@@ -17,6 +17,9 @@
 
 package org.apache.linkis.engineplugin.repl.executor;
 
+import org.apache.linkis.common.io.MetaData;
+import org.apache.linkis.common.io.Record;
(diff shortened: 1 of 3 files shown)


linkis-engineconn-plugins/repl/src/main/java/org/apache/linkis/engineplugin/repl/executor/ReplEngineConnExecutor.java
```diff
@@ -132,14 +143,33 @@ public ExecuteResponse executeLine(EngineExecutionContext engineExecutorContext,
...
     } catch (Exception e) {
       String errorMessage = ExceptionUtils.getStackTrace(e);
       logger.error("Repl engine execute failed : {}", errorMessage);
       engineExecutorContext.appendStdout(LogUtils.generateERROR(errorMessage));
+      return new ErrorExecuteResponse(errorMessage, null);
     }
 
+    String message = outputStream.toString();
+    System.setOut(oldStream);
+    System.out.println(message);
```
peacewong, line 162: should remove this print line?

ChengJie1053: Ok, thanks for reviewing the code

peacewong (approved):
LGTM.

PR #5041, opened 2023-12-14
Fix flink sql kill yarn application and getJobStatus fail

ChengJie1053:
Fix flink sql kill yarn application and getJobStatus fail


peacewong (approved):
LGTM.

PR #5057, opened 2023-12-25
fix: To avoid converting null to "null", skip null passwords

chenmutime:
fix: To avoid converting null to "null", skip null passwords


peacewong (approved):
LGTM.

peacewong:
code need format, just execute cmd: mvn spotless:apply

peacewong (approved):
LGTM.

PR #5058, opened 2023-12-26
Spark etl supports doris

ChengJie1053:
Spark etl supports doris


linkis-engineconn-plugins/spark/pom.xml
```diff
@@ -418,6 +418,12 @@
       <artifactId>kubernetes-model-core</artifactId>
       <version>${kubernetes-client.version}</version>
     </dependency>
+
+    <dependency>
+      <groupId>org.apache.doris</groupId>
+      <artifactId>spark-doris-connector-3.2_2.12</artifactId>
```
peacewong, line 424: Need to be compatible with multiple versions of spark

ChengJie1053: Ok, thanks for reviewing the code

peacewong (approved):
LGTM.

PR #5059, opened 2023-12-27
Fix doris databases sql password could be empty

ChengJie1053:
Fix doris databases sql password could be empty


peacewong (approved):
LGTM.

PR #5170, opened 2024-09-13
Support resultSet direct push for SparkSqlExecutor

ycycse:
Using direct push of result sets, data does not need to be cached on disk and can be calculated in real-time and returned directly from memory. This PR supports direct pushing of resultSets under SparkSqlExecutor.

### DIFF
--- a/linkis-engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/executor/SparkSqlExecutor.scala
+++ b/linkis-engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/executor/SparkSqlExecutor.scala
@@ -118,7 +118,6 @@ class SparkSqlExecutor(sparkEngineSession: SparkEngineSession, id: Long)
 
       if (engineExecutionContext.isEnableDirectPush) {
         submitResultSetIterator(lastTask.getTaskId, df)
-        ReadyForFetchResultResponse()
       } else {
         SQLSession.showDF(
           sparkEngineSession.sparkContext,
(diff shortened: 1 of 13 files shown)


linkis-engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/utils/ArrowUtils.scala
```diff
@@ -0,0 +1,115 @@
...
+          vector
+        case BooleanType =>
+          val vector = new BitVector(field.name, allocator)
+          vector.allocateNew(df.count().toInt)
+          vector
+        case StringType =>
+          val vector = new VarCharVector(field.name, allocator)
+          vector.allocateNew(df.count().toInt)
+          vector
+        case _ => throw new IllegalArgumentException(s"Unsupported data type: ${field.dataType}")
```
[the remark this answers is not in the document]
ycycse: You're right, I will improve this part.

linkis-engineconn-plugins/spark/src/main/scala/org/apache/linkis/engineplugin/spark/executor/SparkSqlExecutor.scala
```diff
@@ -47,6 +44,35 @@ class SparkSqlExecutor(sparkEngineSession: SparkEngineSession, id: Long)
...
+    val dataFrameResponse = DirectPushCache.fetchResultSetOfDataFrame(taskId, fetchSize)
+    if (dataFrameResponse.dataFrame != null) {
+      if (!dataFrameResponse.hasMoreData) {
+        succeedTasks.increase()
+      }
+      FetchResultResponse(
+        hasMoreData = dataFrameResponse.hasMoreData,
+        ArrowUtils.toArrow(dataFrameResponse.dataFrame)
+      )
+    }
```
peacewong, line 72: Else is missing here

ycycse: done. After removing the  job-way to fetch data. This method isn't needed anymore.

peacewong (approved):
LGTM.

PR #5181, opened 2024-09-27
merge 1.7.0 feature

aiceflower:
What is the purpose of the change?
1. The users can specify fixed engine execution
2. Add a new management console to modify the engine status to UnHealthy and automatically kill the engine after task execution
3. Spring upgrade version to 5.3.27
4. Optimize engine reuse logic, large resources cannot reuse small resource engines
5. Optimize the flink log
6. Linkis storage change java code to scala (The core process has some compatibility issues, and some of them have been reversed.)
...


peacewong (approved):
LGTM.

PR #5185, opened 2024-10-06
[BUILD] upgrade CI actions because old versions are no longer supported

pjfanning:
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

They seem to be applying a brown-out period where some builds that need old versions of Node are failing.


peacewong (approved):
LGTM.

PR #5445, opened 2026-06-21
[ISSUE-5130][EC-JDBC] Map NUMERIC columns to decimal

### LINKED ISSUE #5130: [Bug] postgresql dataType Error
### Search before asking

- [X] I searched the [issues](https://github.com/apache/linkis/issues) and found no similar issues.

### Linkis Component

linkis-engineconnn-plugin

### Steps to reproduce

postgresql 创建 数据库表，其中某一字段类型是numeric
执行查询失败，数据类型转换失败
2ef05136e0886f26563b688e128cee4
当我把字段 注释后查询成功
3f9ae0823b311b49f112cd5c3c2c01a

### Expected behavior

数据类型转换 成功

### Your environment

- Linkis version used: 1.4.0
- Environment name and version:
    - jdbc  -4
    - ....

### Anything else

_No response_

- [ ] Yes I am willing to submit a PR!

hutiefang76:
### What is changed?

This PR fixes PostgreSQL numeric column metadata mapping in the JDBC engine plugin. PostgreSQL reports `numeric` columns as `java.sql.Types.NUMERIC`, but `JDBCHelper.getTypeStr` only mapped `Types.DECIMAL` to Linkis `decimal`, leaving `NUMERIC` as `null`.

The value extraction path already handles `Types.NUMERIC` and `Types.DECIMAL` together, so this change applies the same mapping in the type-name conversion path.

### Why?

Fixes #5130.

### Tests

- Reproduced with `mvn -pl linkis-engineconn-plugins/jdbc -am -Dtest=org.apache.linkis.manager.engineplugin.jdbc.executor.JDBCHelperTest -Dsurefire.failIfNoSpecifiedTests=false test`: failed before the fix with `expected: <decimal> but was: <null>`.
- Passed after the fix with the same command.
- Passed JDBC unit test set with `mvn -pl linkis-engineconn-plugins/jdbc -am -Dtest=ConnectionManagerTest,ProgressMonitorTest,JDBCErrorCodeSummaryTest,JDBCHelperTest,JdbcParamUtilsTest,JDBCMultiDatasourceParserTest -Dsurefire.failIfNoSpecifiedTests=false test` (13 tests).

Note: direct `mvn -pl linkis-engineconn-plugins/jdbc test` cannot resolve unpublished Linkis 1.9.0 reactor artifacts from the configured mirror in this local environment, so the verified commands use `-am` to build required modules from source.

### DIFF
--- a/linkis-engineconn-plugins/jdbc/src/main/scala/org/apache/linkis/manager/engineplugin/jdbc/executor/JDBCHelper.java
+++ b/linkis-engineconn-plugins/jdbc/src/main/scala/org/apache/linkis/manager/engineplugin/jdbc/executor/JDBCHelper.java
@@ -145,6 +145,7 @@ public class JDBCHelper {
             case Types.BINARY:
                 retVal = BinaryType.typeName();
                 break;
+            case Types.NUMERIC:
             case Types.DECIMAL:
                 retVal = DecimalType.typeName();
                 break;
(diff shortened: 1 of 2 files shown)


casionone (approved):
LGTM.

PR #5449, opened 2026-06-25
[SECURITY][DS] Complete CVE-2023-49566 fix for non-MySQL JDBC drivers

aiceflower:
## What is the purpose of the change

The CVE-2023-49566 fix shipped earlier only protected the MySQL/StarRocks
`SqlConnection` implementations. The eight other JDBC driver families used by
the metadata-query / datasource-manager modules still streamed user-supplied
`extraParams` straight onto the JDBC URL with no allowlist/denylist, so any
authenticated user could inject driver-specific dangerous properties and reach
`DriverManager.getConnection` with them.

This is the "incomplete fix of CVE-2023-49566" report forwarded by ASF Security
(reporter: greybtw, tested on 1.8.0, CVSS 8.8).

Per-driver sinks reachable before this PR:

| Driver | Sink |
|--------|------|
| PostgreSQL / Greenplum / KingBase | `socketFactory` + `socketFactoryArg` -> reflective class instantiation -> RCE on PG drivers < 42.2.25 / < 42.3.2 |
| DB2 | `clientRerouteServerListJNDIName` -> JNDI injection (the original CVE-2023-49566 sink) |
| Oracle | `oracle.net.tns_admin` / `javax.net.ssl.trustStore` -> TLS/TNS config hijack |
| SQL Server | `jaasConfigurationName` -> JAAS lookup |
| ClickHouse / DM | covered defensively (no known high-risk param today) |

## Brief change log

- **New** `linkis-commons/.../utils/JdbcDriverType.java`: enum identifying the 10 JDBC driver families.
- `linkis-commons/.../utils/SecurityUtils.java`: add generic JDBC security layer
  - `checkJdbcConnParams(JdbcDriverType, host, port, username, password, database, extraParams)` — driver-aware denylist dispatch, URL-decode loop, host-injection guard
  - `buildSecureProperties(JdbcDriverType, username, password, extraParams)` — builds a `Properties` bag with driver-specific force-set security defaults first, then credentials, then user params that don't conflict with the force-set keys
  - Per-driver config keys: `linkis.jdbc.security.check.enable`, `linkis.jdbc.{global,postgres,db2,oracle,sqlserver}.blocked.params`, `linkis.jdbc.{postgres,db2,oracle,sqlserver,clickhouse,dm}.force.params`
- All 16 `SqlConnection` implementations (8 drivers x 2 modules) now:
  1. call `SecurityUtils.checkJdbcConnParams(driverType, ...)`
[...]

PR #5450, opened 2026-06-25
[SECURITY] Block JDBC URL-option injection via DB2 instance field

aiceflower:

Follow-up to #5449 that closes a remaining JDBC URL-option injection sink reported separately against the DB2 `instance` field. The DB2 SqlConnection classes interpolate the user-supplied instance value directly into the URL via `String.format(\"jdbc:db2://%s:%s/%s\", ...)`, so a malicious value like `SAMPLE:traceLevel=1;` becomes `jdbc:db2://host:port/SAMPLE:traceLevel=1;`. The CVE-2023-49566 fix in #5449 denylisted `clientRerouteServerListJNDIName` etc. when they arrive via Properties, but a value smuggled through the URL database segment bypasses Properties entirely and reaches driver options like `traceLevel`/`traceFile`/`traceDirectory`/`traceFileAppend` that allow arbitrary file writes.

Same class of issue affects SQL Server (`;` separator), Oracle (service-name slot), and the `jdbc:mysql://` family (`?`/`&`).

This PR adds four layers of defense:

1. **Runtime validation (primary fix).** New `SecurityUtils.checkDatabaseIsSafe(JdbcDriverType, database)` rejects URL-option separators per driver family. Called from the generic `checkJdbcConnParams` path used by every non-MySQL SqlConnection.
   - DB2 → `: ; ? # &` (matches the advisory's PoC)
   - SQLSERVER → `; ? # &`
   - ORACLE → `? # &`
   - PG/MySQL/ClickHouse/DM/etc → `? # & /`

2. **Expanded DB2 denylist.** `JDBC_DB2_BLOCKED_PARAMS` now also includes `traceLevel, traceFile, traceDirectory, traceFileAppend` so Properties-based injection of the same logging options is blocked too.

3. **Schema-layer validation.** Backfill `value_regex` for the `instance` field of every JDBC data source in `linkis_dml.sql`. `RegExpParameterValidateStrategy` skips validation when `value_regex` is NULL, so the existing schema offered no first-line defense. New regex `^[A-Za-z0-9_.-]+$`.

4. **Upgrade SQL.** Same regex backfilled via `UPDATE ... WHERE key='instance' AND value_regex IS NULL` in the 1.9.0 upgrade script for existing installs.

## Tests

- 6 new unit tests covering: the DB2 PoC (`SAMPLE:traceLevel=1;` and `SAMPLE:traceFile=/tmp/evil;`), the single-character variants for each forbidden char, the SQL Server and MySQL variants, benign-database sanity across all driver families, and the expanded DB2 denylist.
- All 24 tests in `SecurityUtilsTest` pass.

## Relation to prior work

- #5449 closed the CVE-2023-49566 incomplete-fix report by adding the generic JDBC denylist/Properties layer. This PR builds on top of #5449 (same files, same SecurityUtils generic API). If reviewers prefer to land both together, this PR can be rebased onto #5449 instead of master.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/linkis-commons/linkis-common/src/main/java/org/apache/linkis/common/utils/SecurityUtils.java
+++ b/linkis-commons/linkis-common/src/main/java/org/apache/linkis/common/utils/SecurityUtils.java
@@ -118,7 +118,10 @@ public abstract class SecurityUtils {
   private static final CommonVars<String> JDBC_DB2_BLOCKED_PARAMS =
       CommonVars$.MODULE$.apply(
           "linkis.jdbc.db2.blocked.params",
-          "clientRerouteServerListJNDIName,enableSeamlessFailover,JNDIName");
+          // traceLevel / traceFile / traceDirectory / traceFileAppend let the driver write trace
+          // output to arbitrary paths; blocked alongside the JNDI-related options.
+          "clientRerouteServerListJNDIName,enableSeamlessFailover,JNDIName,"
+              + "traceLevel,traceFile,traceDirectory,traceFileAppend");
 
   private static final CommonVars<String> JDBC_ORACLE_BLOCKED_PARAMS =
       CommonVars$.MODULE$.apply(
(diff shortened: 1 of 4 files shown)


casionone (approved):
LGTM.

PR #5458, opened 2026-07-16
[SECURITY] Enforce server-side session map check in ignore-timeout fallback path

aiceflower:
## Summary

`getLoginUserIgnoreTimeout` previously skipped `isTimeoutOrNot` entirely — any ticket that decrypted cleanly (with the known default key) was accepted even if it was never registered via `setLoginUser`. This was the second half of the auth bypass.

This PR adds `isTimeoutOrNot` for all normal users even in the ignore-timeout path. The only exception is `OTHER_SYSTEM_IGNORE_UM_USER` (internal RPC identity via header).

## Changes

- `SSOUtils.getLoginUserIgnoreTimeout`: normal users now go through `isTimeoutOrNot` (session map lookup)
- `OTHER_SYSTEM_IGNORE_UM_USER` remains exempt (internal RPC, not browser-cookie-based)

## Affected files

- `linkis-commons/linkis-module/.../SSOUtils.scala` (+20/-1)

## Deployment impact

| Scenario | Impact |
|---|---|
| HA mode (Redis session sharing) | No impact |
| Non-HA (standalone session maps) | Gateway→backend may fail; switch to HA mode recommended |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/linkis-commons/linkis-module/src/main/scala/org/apache/linkis/server/security/SSOUtils.scala
+++ b/linkis-commons/linkis-module/src/main/scala/org/apache/linkis/server/security/SSOUtils.scala
@@ -17,7 +17,7 @@
 
 package org.apache.linkis.server.security
 
-import org.apache.linkis.common.conf.Configuration
+import org.apache.linkis.common.conf.{CommonVars, Configuration}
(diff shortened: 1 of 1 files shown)


casionone (approved):
LGTM.

